Issue #19, opened 2015-09-16
watson edit fails


braingineer:
```
cogniton@ada ~/research/code/Watson $ watson edit
Traceback (most recent call last):
  File "/home/cogniton/anaconda/bin/watson", line 9, in <module>
    load_entry_point('watson==0.1', 'console_scripts', 'watson')()
  File "/home/cogniton/anaconda/lib/python2.7/site-packages/setuptools-18.1-py2.7.egg/pkg_resources/__init__.py", line 558, in load_entry_point
    
  File "/home/cogniton/anaconda/lib/python2.7/site-packages/setuptools-18.1-py2.7.egg/pkg_resources/__init__.py", line 2682, in load_entry_point
    
  File "/home/cogniton/anaconda/lib/python2.7/site-packages/setuptools-18.1-py2.7.egg/pkg_resources/__init__.py", line 2355, in load
    
  File "/home/cogniton/anaconda/lib/python2.7/site-packages/setuptools-18.1-py2.7.egg/pkg_resources/__init__.py", line 2361, in resolve
    
  File "build/bdist.linux-x86_64/egg/watson/__main__.py", line 6, in <module>
  File "/home/cogniton/anaconda/lib/python2.7/site-packages/click-5.1-py2.7.egg/click/core.py", line 700, in __call__
    return self.main(*args, **kwargs)
  File "/home/cogniton/anaconda/lib/python2.7/site-packages/click-5.1-py2.7.egg/click/core.py", line 680, in main
    rv = self.invoke(ctx)
  File "/home/cogniton/anaconda/lib/python2.7/site-packages/click-5.1-py2.7.egg/click/core.py", line 1027, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/cogniton/anaconda/lib/python2.7/site-packages/click-5.1-py2.7.egg/click/core.py", line 873, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/cogniton/anaconda/lib/python2.7/site-packages/click-5.1-py2.7.egg/click/core.py", line 508, in invoke
    return callback(*args, **kwargs)
  File "/home/cogniton/anaconda/lib/python2.7/site-packages/click-5.1-py2.7.egg/click/decorators.py", line 26, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "build/bdist.linux-x86_64/egg/watson/cli.py", line 481, in edit
[...]

PR #22, opened 2015-09-16
Fix issue #19 (edit command with no frames)

jmaupetit:


watson/cli.py
```diff
@@ -478,6 +478,9 @@ def edit(watson, id):
     The `$EDITOR` environment variable is used to detect your editor.
     """
     if not id:
+        if not len(watson.frames):
+            click.echo("No frame to edit. It's time to start a new one!")
+            return
```
willdurand: Shouldn't we return an exit code here?

jmaupetit: I think click returns 0 when you return from a command, 1 when you raise, and 2 when it fails at parsing arguments. Do you think we should raise?

willdurand: Below, it raises an exception when the frame is not found. It is quite similar..

jmaupetit: You're right. Consistency is the key.

k4nar: Yep, we should raise a `click.ClickException` for that.

jmaupetit:
Done.

PR #24, opened 2015-09-17
User-only install with pip install --user

wehlutyk:
The latter is a very bad idea when the OS provides system-level package management (which includes python packages), and `checkinstall` works painlessly for this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,20 @@ $ pip install td-watson
 
 You might need to run this command as root in order to install Watson globally.
 
+Alternatively, you can choose to install Watson for your user only by running:
+
+```bash
+$ pip install --user td-watson
+```
+
+If after this the `watson` command is not available, you need to add `~/.local/bin/` to your PATH. If your terminal is Bash, you can do this by running:
+
+```bash
+$ echo 'export PATH="$HOME/.local/bin:$PATH"' >> .bashrc  # Add ~/.local/bin/ to your .bashrc PATH
+```
+
+and restarting your terminal session.
+
 ### Development version
 
 The latest development version can be installed using the following commands:


k4nar:
I didn't know about checkinstall, thanks for the heads up!

I just released a package on PyPI and updated the instructions, so there is a conflict with your commit. However your point is still valid, mixing two packaging systems is never a good idea.

I think we could advise to use checkinstall as long as we don't provide any Debian package with our releases. Could you rebase your PR and update your command line to `sudo checkinstall pip install td-watson` ? Of course, that is assuming that checkinstall works with pip flawlessly.

wehlutyk:
Great for the PyPI package!

Just tried `sudo checkisntall pip install td-watson` and it doesn't seem to work (at least, not without tweaking).

But [just discovered](https://stackoverflow.com/questions/7744758/how-to-use-checkinstall-with-python-packages-that-use-setuptools) that with a PyPI package one can `pip install --user td-watson` (so without sudo) and it goes into `~/.local/bin/watson` (which is in `PATH` on default configurations). Would that be a better solution maybe? (The above link explains that checkinstall isn't the best solution for python packages after all.)

I can rebase an PR that

k4nar:
Yeah I guess that `--local` + maybe a message explaining that if the command is not found after that the PATH should be updated would be fine. Nothing specific to Debian though, so I guess you could just update the existing instructions.

wehlutyk:
My git-fu messed things up a bit on my side, but this should be good now :)

wehlutyk:
Side question: `pip install --user td-watson` didn't pull in `arrow` and `Click` (but doing `pip install --user arrow Click` solves this), could that be an issue with the PyPI deps?

willdurand:
Before merging this PR, one has to squash the commits :)

k4nar:
I just checked and `pip install --user td-watson` did install `click` and `arrow`. Maybe you already had them.

wehlutyk:
Squashed. Thanks @willdurand!

@k4nar you're right about that (tried again and it works), I don't what happened on my side then.

README.md
```diff
@@ -28,6 +28,20 @@ $ pip install td-watson
 
 You might need to run this command as root in order to install Watson globally.
 
+Alternatively, you can choose to install Watson for you user only by running:
```
willdurand: should be "for you**r** user", right?

wehlutyk: Thanks, fixed

willdurand:
:+1:

k4nar:
Awesome, thanks!

PR #35, opened 2015-10-19
Update README - bashrc

mms-segu:
Make the bashrc-update command copy-pastable.
Also add source command to activate the changes.


willdurand:
thank you!

PR #36, opened 2015-10-19
Add configurable date and time to output of `status` command

### LINKED ISSUE #33: status should also report the exact start date/time
Information like `an hour ago` may be pretty but often isn't very useful or accurate.

Currently you have to manually open the `state` file any then convert the unix timestamp to see the exact start time of a running frame. Or did I miss another way to do this?

SpotlightKid:
Fixes #33. See also the discussion there.

In this version I uses `Arrow.strftime()` to format the date and time, so arbitrary additional fixed strings can be included, e.g. `Am %d.%m.%Y um %H:%M Uhr`.

Since `SafeConfigParser` (resp. `ConfigParser` in Python 3) uses the percent sign for interpolation too and I don't think this feature is needed, I included a change to the import of the modul so that the version without interpolation is used. With this change you don't have to use two percent signs for the date/time placeholders in the config file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,11 +84,23 @@ be recorded.
 
 ### status
 
-Display the time spent since the current project was started.
+Display when the current project was started and the time spent since.
+
+You can configure how the date and time of when the project was started are
+displayed by setting `options.date_format` and `options.time_format` in the
+configuration. The syntax of these formatting strings and the supported
+placeholders are the same as for the `strftime` method of Python's
+[datetime.datetime][datetime] class.
+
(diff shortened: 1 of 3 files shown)


willdurand:
Hi! Thanks for your PR :)

Could you please:

1. update the doc to introduce your new feature?
2. squash your commits?

Thank you!

SpotlightKid:
Done.

willdurand:
Looks good to me, :+1:

jmaupetit:
:+1: Nice! Thx!

willdurand:
thanks @SpotlightKid

PR #37, opened 2015-10-19
Add configuration for Bash command line completion

### LINKED ISSUE #1: [feature] shell completion
Add completion on commands project names and tags may be useful.

SpotlightKid:
Fixes #1 

I've also added a new command `frames` to list all frame IDs to support completion for the `edit` and `remove` commands. I've put this into an extra commit, so you can choose to include it or not.

Also added some documentation for it in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,15 @@ $ cd Watson/
 $ python setup.py install
 ```
 
+### Command line completion
+
+If you use a Bash-compatible shell, you can install the `watson.completion`
+file from the source distribution as `/etc/bash.completion.d/watson` - or
+wherever your distribution keeps the Bash completion configuration files.
+After you restart your shell, you can then just type `watson` on your command
+line and then hit TAB to see all available commands. Depending on your input,
+it completes watson commands, command options, projects, tags and frame IDs.
+
 ## Commands
 
 Here is the listing of all the commands available with Watson. You can also
@@ -188,6 +197,19 @@ Thursday 17 April 2014
 Wednesday 16 April 2014
         02cb269  09:53 to 12:43  2h 50m 07s  apollo11  [wheels]
         1070ddb  13:48 to 16:17  2h 29m 11s  voyager1  [antenna, sensors]
+```
+
+### frames
+Display the list of all frame IDs.
+
+This is mainly useful for implementing Bash command line completion.
+
+```
+$ watson frames
+f1c4815
+9d1a989
+8801ec3
+[...]
 ```
 
 ### projects
--- a/watson.completion
+++ b/watson.completion
@@ -1,0 +1,66 @@
+# Bash completion support for watson
+#
+# Install this as `/etc/bash_completion.d/watson` and restart your shell.
+
+_watson_complete () {
+  local cur cmd commands frames projects tags
+
+  commands='cancel config edit frames help log projects remove report start status stop sync tags'
+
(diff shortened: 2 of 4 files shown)


watson.completion
```diff
@@ -0,0 +1,66 @@
+# Bash completion suppoort for watson
```
willdurand: s/suppoort/suport

README.md
```diff
@@ -52,6 +52,15 @@ $ cd Watson/
...
 ```
 
+### Command line completion
+
+If you use a Bash-compatible shell, you can install the `watson.completion`
+file from the source distribution as `/etc/bash.completion.d/watson` - or
+wherever your distribution keeps the Bash completion configuration files.
+After you restart your shell, you can then just type `watson` on your command
+line and then hit TAB to see all available commands. Depending on your input,
+it completes watson commands, command options, projects, tags and frame IDs
```
willdurand: Period is missing at the end of this sentence.

willdurand:
Could you please pass [shellcheck](https://github.com/koalaman/shellcheck) on the script file?

SpotlightKid:
I ran `shellcheck watson.completion`, but the complains it has do not make sense to me:

```
In watson.completion line 1:
# Bash completion support for watson
^-- SC2148: Tips depend on target shell and yours is unknown. Add a shebang.

In watson.completion line 19:
          COMPREPLY=($(compgen -W "-e --edit" -- ${cur})) ;;
                                                 ^-- SC2086: Double quote to prevent globbing and word splitting.
[a lot more of those...]
```

1. This is not a script, but a something to be sourced by `/etc/bashrc`, therefor it has no shebang line.
2. `${cur}` must not be quoted, because otherwise `COMPREPLY` is an array.

willdurand:
Thank you! Looks good to me, so :+1:

jmaupetit:
@SpotlightKid Thank you for this!

Commands and projects completions work just fine, but tags completion does not :( I always have `+0` as completion... 

FYI my bash release is `GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin15)`

SpotlightKid:
Hmm, maybe your version of `sed` works differently than mine. Look at line 33 or 55 where I use `sed` too add a plus sign in front of every line returned by `watson tags`, by using the placeholder `\0` for the entire regular expression match. Maybe your sed doesn't support this placeholder?

Can you replace `sed 's/.*/+\0/'` with `awk '$0="+"$0'` in both places and see if that works for you?

willdurand:
@SpotlightKid don't you have to use a group for that? (i.e. adding parenthesis)

SpotlightKid:
In fact, I just noticed that adding plus signs is not needed when completing the argument of the `-T` option to `log` and `report` (line 33). Amended the commit.

SpotlightKid:
@willdurand `\0` normally is the entire match. It works on my system (`sed (GNU sed) 4.2.2`). However, the awk solution should work better, so I used this and squashed the commits again.

willdurand:
cool! BTW, what's your OS? Because we run OSX here, and there are differences between the unix tools..

SpotlightKid:
Manjaro (Arch Linux derivate)

jmaupetit:
@SpotlightKid the `awk` solution works like a charm on OSX!

Thank you so much for this!

PR #39, opened 2015-10-20
Keep cjk characters unescaped.

bcho:
Instead of showing escaped characters in editing mode:

```json
{
    "project": "a-cool-project",
    "start": "2015-10-19 13:27:11",
    "stop": "2015-10-19 14:53:08",
    "tags": [
        "\u91cd\u6784"
    ]
}
```

Escaped characters should be:

```json
{
    "project": "a-cool-project",
    "start": "2015-10-19 13:27:11",
    "stop": "2015-10-19 14:53:08",
    "tags": [
        "重构"
    ]
}
```

Which can give a better look.

### DIFF
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -353,6 +353,9 @@ def test_save_current_without_tags(watson):
             assert isinstance(result['start'], (int, float))
             assert result['tags'] == []
 
+            dump_args = json_mock.call_args[1]
+            assert dump_args['ensure_ascii'] is False
+
 
 def test_save_empty_current(config_dir):
     watson = Watson(current={'project': 'foo', 'start': 0},
@@ -410,6 +413,9 @@ def test_save_changed_frame(config_dir):
             assert len(result) == 1
             assert result[0][2] == 'bar'
             assert result[0][4] == ['A', 'B']
+
+            dump_args = json_mock.call_args[1]
+            assert dump_args['ensure_ascii'] is False
 
 
 def test_save_config_no_changes(watson):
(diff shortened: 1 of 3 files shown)


bcho:
@TailorDev Should we also unescaped the cjk characters when saving the frames? https://github.com/TailorDev/Watson/blob/develop/watson/watson.py#L144

jmaupetit:
@bcho thank you for this!
@k4nar WDYT?

jmaupetit:
@bcho I think you are right: we should unescape cjk characters when saving. Can you add this to this PR?

bcho:
@jmaupetit Patch updated!

willdurand:
Do we have tests for that?

bcho:
@willdurand nope, I think this should be tested by hand.

k4nar:
:+1: 

I think we should have a test case `save_frame_with_cjk_char` though. You will find all the tests for the `save` function here: https://github.com/TailorDev/Watson/blob/develop/tests/test_watson.py#L319

bcho:
@k4nar @willdurand ok, I will submit test later :)

tests/test_watson.py
```diff
@@ -353,6 +353,9 @@ def test_save_current_without_tags(watson):
             assert isinstance(result['start'], (int, float))
             assert result['tags'] == []
 
+            dump_args = json_mock.call_args[1]
+            assert dump_args['ensure_ascii'] is False
```
bcho: @k4nar @willdurand Is it ok to test calling params only? I don't figure out how to check the dumped json value.

k4nar: Yeah, as we mock the json call we can't really check the output.

k4nar:
LGTM, the test case could be more thorough but we were already UTF-8 compliant, so I guess it's fine.

jmaupetit:
:+1:

Issue #31, opened 2015-10-17
Include (or provide an option) to include current frame in reports


waynew:
Right now the current frame appears to be excluded from reports. If you have no frames, for instance, and you do this:

    $ watson start fnord
    $ sleep 10; # Just so there's some time
    $ watson report

You'll see that watson does not display anything. I think either it should by default, or you could do something like:

    $ watson report --with-current

Most of the time I want to see how much time I've spent on a project including what I've been working on for the past 3.5 hours. Right now I think the only option is to do something like this:

    $ watson report
    $ watson status

And then manually combine the two numbers. There must be a better way :)

jmaupetit:
Interesting scenario and interesting proposed solution. I give it a :+1:

SpotlightKid:
+1 (but please also provide a short form of the option!)

k4nar:
Shouldn't we have a config option for that? It looks like a personal preference to me.

jmaupetit:
An option is a good idea for defaults, but it must be overridable when calling the `report` command.

SpotlightKid:
A sidenote: as more options are used, it would be good thing to provide a sane wrapper around `ConfigParser` so you don't have to do stupid things like this all the time:

```
option = (watson.config.get('section', 'option')
          if watson.config.has_option('section', 'option') else default)
```

Even if it's only wrapping the above code in sub-classes `get()` method:

```
class SaneConfigParser(ConfigParser):
    def get(self, section, option, default=None):
        return (self.get(section, option)
                if self.has_option(section, option) else default)
```

PR #43, opened 2015-10-22
Add wrapper for `RawConfigParser` to make option access more convenient

SpotlightKid:
See comment on #31 [here](https://github.com/TailorDev/Watson/issues/31#issuecomment-150196100).

### DIFF
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -16,7 +16,7 @@ import requests
 import arrow
 
 from watson import Watson, WatsonError
-from watson.watson import ConfigParser
+from watson.watson import ConfigurationError, ConfigParser
 
 PY2 = sys.version_info[0] == 2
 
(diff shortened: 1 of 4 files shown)


SpotlightKid:
Tests fail on Python 2.7 because `RawConfigParser` is an old-style class there, so `super` can't be used. Fixing it now.

SpotlightKid:
Any thoughts on this?

k4nar:
LGTM

waynew:
:+1: from me (FWIW). My PR #47 depends on this implementation (or the actual config value to be present).

jmaupetit:
:+1:

SpotlightKid:
While looking at #47, I noticed one minor issue. The `ConfigParser` wrapper breaks the `getint`, `getfloat` and `getboolean` methods from the base class. Two solutions:

1. Rename `get` to sth like `getoption`, so it doesn't conflict.
2. Wrap the above mentioned methods as well.

Seems like a lot of trouble, but it would allow to support boolean options like this:

```
[options]
auto_stop = yes
report_current_frame = 0

if config.getboolean('options', 'auto-stop')
    ...
```

I tried to make wrappers that delegate to the base class but it was impossible to do that in a sane way that works in Python 2 and 3. So I basically re-implemented them:

```
class ConfigParser(RawConfigParser):
    """A simple wrapper for RawConfigParser to make options access easier."""

    def get(self, section, option, default=None, **kwargs):
        """
        Return value of option in given configuration section as a string.

        If option is not set, return default instead (defaults to None).

        """
        return (RawConfigParser.get(self, section, option, **kwargs)
                if self.has_option(section, option) else default)

    def getint(self, section, option, default=None):
        val = self.get(section, option, default)
        return val if val is None else int(val)

    def getfloat(self, section, option, default=None):
        val = self.get(section, option, default)
        return val if val is None else float(val)

    def getboolean(self, section, option, default=False):
        val = self.get(section, option)
        return default if val is None else val.lower() in (
            '1', 'on', 'true', 'yes')
   ...
```

I also passed `kwargs` to `RawConfigParser.get()` for more compatibility with Python 3's `configparser`.
Should I update my PR with this?

SpotlightKid:
Moved `ConfigParser` into its own sub-module, because it was getting to long. Added the `getint`, `getfloat` and `getboolean` methods (slightly different than the code above!) with docstrings and tests.

jmaupetit:
:+1:
@SpotlightKid thank you again for your active work.
@k4nar let's merge this!

k4nar:
:tada:

PR #47, opened 2015-10-29
Stop current frame when starting something else

waynew:
I don't know if it's just the way I work, but I dislike having to say

    $ watson stop
    $ watson start <new project> [new tags]

So I now I don't have to :)

### DIFF
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -119,16 +119,17 @@ def start(watson, args):
     $ watson start apollo11 +module +brakes
     Starting apollo11 [module, brakes] at 16:34
     """
-    try:
-        frame = watson.stop()
-        click.echo("Stopping project {} {}, started {}. (id: {})".format(
-            style('project', frame.project),
-            style('tags', frame.tags),
-            style('time', frame.start.humanize()),
-            style('short_id', frame.id)
-        ))
-    except WatsonCliError:
-        pass
+    if watson.config.get('options', 'auto_stop'):
+        try:
+            frame = watson.stop()
+            click.echo("Stopping project {} {}, started {}. (id: {})".format(
+                style('project', frame.project),
+                style('tags', frame.tags),
+                style('time', frame.start.humanize()),
+                style('short_id', frame.id)
+            ))
+        except WatsonCliError:
+            pass
 
     project = ' '.join(
         itertools.takewhile(lambda s: not s.startswith('+'), args)
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -119,7 +119,7 @@ def start(watson, args):
     $ watson start apollo11 +module +brakes
     Starting apollo11 [module, brakes] at 16:34
     """
-    if watson.config.get('options', 'auto_stop'):
+    if watson.config.get('options', 'stop_on_start'):
         try:
             frame = watson.stop()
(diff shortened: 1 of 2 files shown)


jmaupetit:
I am not sure this must be the default behavior. An option or global configuration must be set for that.

k4nar:
@jmaupetit : :+1:

waynew:
@jmaupetit that certainly makes sense. I don't see any examples in cli.py that actually *use* any kind of config (except for the config, which has to load up the config first), so I don't think I have any guidance to go on for how to approach that...

SpotlightKid:
@waynew  Look at PR 36 and PR 43

waynew:
That's #36 and #43 for the lazy :)

waynew:
Thanks @SpotlightKid.

---

Updated to use the config, but this *does* depend on #43 being merged in or it will break, absent that config in the file.

I used the setting `auto_stop` because it was the first and most accurate setting name that came to mind that wasn't ridiculously long like `stop_current_task_when_starting_another`. I don't have strong feelings about the former if there's something else concise and maybe more descriptive.

jmaupetit:
What about `stop_on_start`?

waynew:
I need to change this to use `.getboolean`, as #43 will parse `stop_on_start = False` correctly. I think how it is that would probably just evaluate to `'False'`, which is `True` :trollface:

k4nar:
This should be rebased now that #43 is merged.

waynew:
@k4nar done

Also did the `.get` -> `.getboolean` thing

jmaupetit:
:+1:

k4nar:
@waynew : Great, thanks !

Could you update the docstring and the readme to document this new option ?

watson/cli.py
```diff
@@ -119,6 +119,18 @@ def start(watson, args):
     $ watson start apollo11 +module +brakes
     Starting apollo11 [module, brakes] at 16:34
     """
+    if watson.config.getboolean('options', 'stop_on_start'):
```
k4nar: We should check `watson.is_started` instead of using a try/except.

SpotlightKid:
@waynew FWIW, since I included the feature in #49 as well, I added documentation to the README and `cli.start` there already.

waynew:
@k4nar given that @SpotlightKid already added some docs, should I just copy those? Not sure if there's a reasonable way to get that particular commit in my branch :P

k4nar:
@waynew : Yes you can copy the docs. When @SpotlightKid will rebase, it will no longer be in his commit :) .

jmaupetit:
@k4nar Ready to be merged?

k4nar:
@jmaupetit : I would like to see some docs with that PR.

SpotlightKid:
@waynew Did you have time to include the docs yet? I want to present Watson at our local user group meeting next week (Wed) and it would be cool if this PR and #49 could be merged by then!

waynew:
@SpotlightKid / @k4nar / @jmaupetit Docs :+1:

SpotlightKid:
@waynew Thanks!

k4nar:
Great, thank you for your time :+1:!

jmaupetit:
@waynew you rock! :+1:

jmaupetit:
@SpotlightKid Please let us know if you have any feedback from your talk!

PR #50, opened 2015-11-04
Add flake8 to test environments to always run

SpotlightKid:
So flake8 is always run locally, when you check our commits with tox, same as travis does and you don't get any surprises.


k4nar:
Good idea :+1: !

willdurand:
:+1:

PR #52, opened 2015-11-10
Current work-week shortcut & progress bar

waynew:
Not sure if anyone else is interested in this feature, but most of the time I'm interested in what I've done this current workweek (as opposed to the last week).

So I've added a shortcut to `report` that lets you do that with 3 extra keystrokes: ` -w`.

Many times I'm *also* interested if I've put in enough time this week. Sure I can look at the numbers and make a comparison, but a chart is easier to grok and since Click just had this one sitting around... why not? :trollface: 

This also adds a new option:

```
[options]
target_hours_per_week = 40  #unless you're a workaholic ;)
```

**With that option:**

```
⚘ watson report -w                    
Mon 09 November 2015 -> Fri 13 November 2015

watson - 45m 42s
        [personal         45m 42s]
        [progress bar     45m 42s]

Work completed: 0.8h/22.9h  [#-----------------------------------]    3%
```

**And without:**

```
⚘ watson report -w                   
Mon 09 November 2015 -> Fri 13 November 2015

watson - 45m 42s
        [personal         45m 42s]
        [progress bar     45m 42s]
```

---
## Possible Improvements

If this gets merged in, an idea for future enhancement would be the ability to have per-tag target hours in the config, e.g.

```
[options]
target_hours_per_week = 40

[tag:watson]
target_hours_per_week = 5
```

It could also be nice to provide alternate `workweek` config options for those with alternative work schedules, e.g.

```
[options]
workweek_start = Tuesday
workweek_end = Saturday
```

Or maybe even:

```
[options]
work_days = Monday, Wednesday, Friday
```

I don't really have a need for any of those options at the moment, but they might be useful to others.

### DIFF
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from __future__ import division
 
 import json
 import datetime
@@ -247,8 +248,11 @@ def status(watson):
               help="Reports activity only for frames containing the given "
               "tag. You can add several tags by using this option multiple "
               "times")
+@click.option('-w', '--workweek', 'special', flag_value='workweek',
+              help="Shortcut for the current work week, overriding -f and -t "
+              "by setting them to Monday and Friday respectively.")
 @click.pass_obj
-def report(watson, from_, to, projects, tags):
+def report(watson, from_, to, projects, tags, special):
     """
     Display a report of the time spent on each project.
 
@@ -312,6 +316,12 @@ def report(watson, from_, to, projects, tags):
             [steering 10h 33m 37s]
             [wheels   10h 11m 35s]
     """
+    if special == 'workweek':
+        now = arrow.now()
+        monday = now - datetime.timedelta(days=now.isoweekday()-1)
+        friday = monday + datetime.timedelta(days=4)
+        from_, to = monday, friday
+
     if from_ > to:
         raise click.ClickException("'from' must be anterior to 'to'")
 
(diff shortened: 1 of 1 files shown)


tomleo:
I really like this idea :+1:

watson/cli.py
```diff
@@ -371,6 +381,23 @@ def report(watson, from_, to, projects, tags):
             style('time', '{}'.format(format_timedelta(total)))
         ))
 
+    target_hours_per_week = watson.config.getfloat('options',
+                                                   'target_hours_per_week')
+    if target_hours_per_week is not None:
+        sec_per_min = 60
+        min_per_hr = 60
+        offset = 1000
+        weeks = abs(span.start - span.stop).days/7
```
waynew: I just realized that this calculation isn't quite right. This works if there are 7 days in the workweek, but a typical week is 5 days.

larsborn:
if no one opposes, I would take this one over. Or did it fell off the feature list entirely?

SpotlightKid:
At this point I would suggest starting a new separate `watson-report` program and invest a little thought in how the command line interface can designed user-friendly and still with room for additions.

larsborn:
so you suggest to not integrate such features into watson but write a separate program (that uses watson for example)?

k4nar:
I don't know if we need an external command for that. Maybe a clever addons system could cover that.

SpotlightKid:
The program could still live in the `watson` package. but I would add a separate module and command line entry point in `setup.py`.

jmaupetit:
I think it's a good idea to decrease sub-commands complexity. I invite all of you to up/down vote previous @SpotlightKid comment to see where we are going on this.

larsborn:
are there already more then this one "more advanced" report feature planed? 

What insistently springs to my mind is a report, that does the following for every "work day" (definition outstanding): subtract the numbers of hours worked from a configurable "target number" of hours. Then sum up these differences. This would enable easy overtime tracking. This is somehow related to this PR but also a bit different imho.

SpotlightKid:
Other features I would like to see integrated / moved into a watson report command are:

* output in CSV or JSON format
* long form output with log messages (from pending PR #119)
* configurable rounding (#109)

larsborn:
I feel, that we totally lost the "next action" in this thread. Since I am the new kid here, I'm totally happy with just following orders, so if someone tells me to put something into an external command, I'll do it ;-)

Sooo, what should we do?

SpotlightKid:
* Read the [click documentation](http://click.pocoo.org/).
* Copy `watson/cliy.py` to `watson/report.py`.
* Throw out everything not related to the report command.
* Set up a setuptools console command entry point in `setup.py` for a command named `watson-report` or similar, which calls into this file.
* Design a command line API for the new command that integrates the existing functionality of `watson report` but gives a good foundation to add more report types and functions. Feel free to rename options and/or restructure arguments to achieve better consistency.
* Make a WIP PR for discussion.

jmaupetit:
Thx for taking the lead on this @SpotlightKid !

Issue #28, opened 2015-10-12
Start a frame with a defined date and time


lunika:
Hi,

Most of time I forget to start watson when I start a new task. If I start watson 30 minutes after I start my task, I have to wait the end of the task and then edit the frame and change the date. 

It would be great to have an option like `--date` where you can specify the date and time.

WDYT ?

jmaupetit:
Thank you for your feedback dude :smile_cat:

I have encountered the same situation quite often and I think that an alternative to the `--date` option to the `start` command would be to be able to edit the current session via:

    $ watson edit --current

It would be easier than providing a date time via command line.

@k4nar WDYT?

k4nar:
Yeah, an option to specify the time at which the frame should have started would be cool (a time in the future being illegal).

I don't really see a point to require a full date though, something like `HH:MM` or eventually `HH:MM:SS` would be enough (it could be tricky around midnight but you can still do this the old way and edit the frame).

What would be a good name for that option ? Maybe `-f, --from`.

lunika:
If you use the [iso 8601](https://fr.wikipedia.org/wiki/ISO_8601) notation you don't care if it's a date or a time.

jmaupetit:
@lunika good idea, if we were computers. But we are not :)
@k4nar doing this "the old way" by editing the `state` file (current frame) is not human friendly as the start datetime is a timestamp.

Moreover, the start -f does not cover the scenario where you forgot tags or made a typo in the project name. Do you really prefer this?

```bash
# Oops I forgot to launch watson 30 minutes ago
$ watson start foo -f HH:MM
# Oops I forgot the proper tag
$ watson stop
$ watson delete FRAMEID
$ watson start foo +tag -f HH:MM
```

To this?

```bash
# Oops I forgot to launch watson 30 minutes ago
$ watson start foo
$ watson edit --current
# Add tags, fix start field, etc.
```

k4nar:
Maybe `watson edit` could edit the current frame by default. Right now it edits the last recorded one.

jmaupetit:
:+1:

lunika:
:+1:

waynew:
I can give a :+1: to editing the current frame. That's basically what I think I want to do.

SpotlightKid:
Yes, if there is a running frame, `edit` should to default to this one, instead of the last finished one, IMHO.

willdurand:
:+1:

PR #54, opened 2015-11-13
edit defaults to current frame (see #28)

SpotlightKid:
Also allow to pass a frame index instead of a frame id to `watson edit`.


jmaupetit:
Yay! Thank you for this! :+1:

willdurand:
:+1:

PR #55, opened 2015-11-14
Add option to display program version (fixes #51)

### LINKED ISSUE #51: Display version on the command line

SpotlightKid:
Moves the version number from `setup.py` into `watson/version.py`, from where it is imported by `watson/watson.py` and read via `exec` by `setup.py`. Exec is used, so that the version number isn't accidentally imported from an installed version of Watson. The version is in a separate module, so `setup.py` doesn't have to import/exec `watson/watson.py`, which would trigger dependencies. The mechanism can also be used to move additional package metadata - such as author, license, etc. - into `version.py` (it should then probably be renamed).

I used `"Watson"` (capitalized) as the program name, when displayed by click. I hope that's right, but it can easily be changed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,12 +1,23 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""Setup file for the Watson distribution."""
+
+from os.path import join
+
 from setuptools import setup
 
 with open('README.rst') as f:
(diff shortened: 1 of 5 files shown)


willdurand:
:+1:

jmaupetit:
:+1:

PR #45, opened 2015-10-22
Python 3.5 support - Fix #44

k4nar:
cf #44

PR #56, opened 2015-11-14
Add Python 3.5 to tox and travis config (fixes #44)

### LINKED ISSUE #44: Python 3.5 env should be added to tox.ini
Arch Linux already has Python 3.5 as the default Python version now.

SpotlightKid:
Run tox on travis via Python 3.5.

Fixes #44. Obsoletes #45.

The problem with Travis mentioned in #45 is that it does not find Python 3.5, when using the Python 2.7 environment. Since we use tox to run tests for different Python versions, the simple solution was to run tox always via Python 3.5. The magic trick with the build environment matrix mentioned [here](https://github.com/travis-ci/travis-ci/issues/4794#issuecomment-143568134) isn't needed in this case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,12 @@
 language: python
 python:
-  - "2.7"
+  - "3.5"
 env:
   - TOX_ENV=flake8
   - TOX_ENV=py27
   - TOX_ENV=py33
   - TOX_ENV=py34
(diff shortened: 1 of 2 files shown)


willdurand:
:+1:

jmaupetit:
:+1:

PR #61, opened 2015-12-01
Adding the `merge` command #18

### LINKED ISSUE #18: Add merge command
I am using dropbox with [Mackup](https://github.com/lra/mackup) to sync watson `frames` between all my machines. In some cases, a conflict file is generated due to a missing sync when one of my machine was used off-line. Hence, it would be nice to be able to merge two conflicting watson `frames` files.

Usage may be:

```bash
$ watson merge frames frames-with-conflicts
```

By default, `merge` will output statistics about the merge operation, like:

```bash
$ watson merge frames frames-with-conflicts
120 frames will be left unchanged
12  frames will be merged
3   frame conflicts need to be resolved
```

To perform the merge operation, there is an option for that:

```bash
$ watson merge frames frames-with-conflicts --no-dry-run
120 frames unchanged
12 frames merged
Will resolve conflicts:
frame 8804872:
< {
<     "project": "tailordev",
<     "start": "2015-07-28 09:33:33",
<     "stop": "2015-07-28 10:39:36",
<     "tags": [
<         "intern",
<         "daily-meeting"
<     ]
< }
---
> {
>     "project": "tailordev",
>     "start": "2015-07-28 09:33:33",
>     "stop": "**2015-07-28 11:39:36**",
>     "tags": [
>         "intern",
>         "daily-meeting"
>     ]
> }
Select the frame you want to keep: left or right? (L/r)
...
```

I am widely open to discussion & suggestions.

EmilMaric:
Resolves #18 

This pull request is to add the `merge` command for conflicting frames files. For a first run, I decided to address the common case of conflicting frames with the same UUIDs and differing attributes.

To invoke the command, you can run
```bash
$ watson merge FRAMES FRAMES_WITH_CONFLICT [--no-dry-run]
```

I also added a test case for the merge statistics (not the actual merging functionality since this will be addressed in issue #5). The test case relies on a sample `frames` file and a conflicting `frames` file, which is located in `tests/resources/`.

I didn't add anything to the `README` or any bash completion, since I wasn't sure if you guys wanted to add that, or if I should be adding that.


k4nar:
Thank you very much for your contribution. Now we'll take some time to review it :) .

watson/cli.py
```diff
@@ -822,3 +822,111 @@ def sync(watson):
...
+    dig = len(str(max(len(original_frames), len(merging), len(conflicting))))
+
+    click.echo("{:<{width}} frames will be left unchanged".format(
+        len(original_frames) - len(conflicting), width=dig))
+    click.echo("{:<{width}} frames will be merged".format(
+        len(merging), width=dig))
+    click.echo("{:<{width}} frames will need to be resolved".format(
+        len(conflicting), width=dig))
+
+    if not no_dry_run:
```
k4nar: I'm not really sure to like this behavior. I would rather have a `Do you want to continue ? [Y/n]` question. Forcing the user to launch the command again with a new option is a bit unexpected.
Click already has a helper for asking this type of question (it's used in Watson [here](https://github.com/EmilMaric/Watson/blob/feature/merge/watson/cli.py#L716-L725).

watson/cli.py
```diff
@@ -822,3 +822,111 @@ def sync(watson):
...
+            original_frames[conflict_frame.id] = conflict_frame
+
+    # merge in any non-conflicting frames
+    for frame in merging:
+        start, stop, project, id, tags, updated_at = frame.dump()
+        original_frames.add(project, start, stop, tags=tags, id=id,
+                            updated_at=updated_at)
+
+    # overwrite the original frames file if any frames have changed
+    if original_frames is not None and original_frames.changed:
```
k4nar: I don't think we should touch the JSON file in `cli.py`. Instead I think we should use `watson.frames = original_frames` and then `watson.save()`.

EmilMaric: Since watson.save() will use the default frames file, does the user really need to supply the path to the original frames file? Can we then just supply the path to the confict file and assume that the user wants to merge it into the default frames file?

We had originally based the design off of the discussion in issue #18 where the design was to specify both path to files. However, if we choose to use the default frames file, we may no longer need the first argument.

k4nar: Indeed, I think the user should only give the conflicting path.

watson/cli.py
kwou: I agree that this be better for a user's workflow. From the comments in issue #18, it looks like there is a desired option for the `--no-dry-run`. 

Would this prompt for the user to continue would be used only for the case where the user did not specify the `--no-dry-run` option?

k4nar: In my opinion by default the command should ask a user, and we could  have a `--force` option which would be the equivalent of answering yes.

@jmaupetit : WDYT ?

jmaupetit: Even if this diff is outdated, I do agree.

watson/cli.py
jmaupetit: Despite the aforementioned issue #18 usage, I think you are right about this: providing only the conflicting path would be sufficient. :+1:

jmaupetit:
Well, thank you so much for this @EmilMaric @kwou and @deb0324! I will test this asap and give you relevant feedback.

jmaupetit:
@EmilMaric when this PR will be ready (and I am pretty sure it will be soon), a final polish with an updated README / bash completion would be really nice :heart:

watson/cli.py
EmilMaric: Thanks, we've changed this so that the `--force` option is available and removed the `--no-dry-run` option.

watson/cli.py
EmilMaric: Thanks, we've changed this so that the `--force` option is available and removed the `--no-dry-run` option.

EmilMaric:
The README and bash completion have now been added!

SpotlightKid:
1. Can we have a short form (`-f`) of the `--force` option as well?

2. I suggest you add the following to `watson.completion` as well:

```
diff --git a/watson.completion b/watson.completion
index c1edd9f..fccf896 100644
--- a/watson.completion
+++ b/watson.completion
@@ -37,6 +37,15 @@ _watson_complete () {
               COMPREPLY=($(compgen -W "-f -t -p -T --from --to --project --tag" -- ${cur})) ;;
           esac
           ;;
+        merge)
+          case "${cur}" in
+            -*)
+              COMPREPLY=($(compgen -W "-f --force" -- ${cur})) ;;
+            *)
+              COMPREPLY=($(compgen -o filenames -f -- ${cur}))
+              ;;
+          esac
+          ;;
         remove)
           case "${cur}" in
             -*)
```

kwou:
@SpotlightKid Thanks for the suggestions! We've made the changes.

README.rst
```diff
@@ -359,6 +359,64 @@ The URL of the server and the User Token must be defined in your
...
+merge
+~~~~~
+
+Perform a merge of the frame file, with a conflicting frames file.
+
+When storing the frames on a file hosting service, there is the
+possibility that the frame file goes out-of-sync due to one or more of the
+connected clients going offline. This can cause the frames to diverge.
+
+If the ``--force`` command is not specified, only the statistics of the
```
k4nar: I believe this is not true anymore :) .

README.rst
```diff
@@ -359,6 +359,64 @@ The URL of the server and the User Token must be defined in your
...
+merge will be shown, and the merge will not be performed.
+
+The argument ``FRAMES_WITH_CONFLICT`` is a path to the the conflicting
```
k4nar: I think it can be a bit confusing to name the argument here. I guess `The only argument` would be fine.

watson/utils.py
```diff
@@ -32,3 +34,17 @@ def sorted_groupby(iterator, key, reverse=False):
...
+    """
+    Wrapper for the `value_proc` field in `click.prompt`, which validates
+    that the user response is part of the list of accepted responses.
+    """
+    def value_proc(user_input):
+        if user_input in opt_list:
+            return user_input
+        else:
+            raise UsageError("Response should be one of [{}]".format(
+                ','.join([str(x) for x in opt_list])))
```
k4nar: This is not important for this PR, but here is a tip in Python:
You don't actually need to give `join` a list comprehension. You can remove the `[` and `]`, without them it becomes a "generator expression" and `join` will iterate over it. Cf [this PEP](https://www.python.org/dev/peps/pep-0289/) for more info.

k4nar:
This PR looks more and more awesome. Great job!

I'm a bit worried about timezones though. We add several issues with them, and I always struggle to get my head around it. @EmilMaric @kwou : Did you made your manual tests using a timezone different from UTC ?

README.rst
kwou: Good catch! Thanks for this, I'll be sure to update the README.

watson/utils.py
EmilMaric: Thanks! I remember spotting this earlier, but I thought I had changed `join()` to use the generator expression form, but evidently I forgot. It wouldn't hurt to make this change, and it'll make this line look a little more readable.

kwou:
@k4nar We did consider the possibility of problems with the timezones. However, we found that the frames file always stored the times in UTC and whenever we loaded the file it would convert it to the local timezone. This would mean that for any comparisons we made for the merge operations, the frames would always be in the same timezone. So we should not have a problem. Is this the issue you had in mind?

k4nar:
@kwou : Great, thank you! I remember we add issues with that with the `edit` command (converting timezones back and forth).

jmaupetit:
@kwou @EmilMaric @SpotlightKid @deb0324 Thank you all for this!

PR #63, opened 2015-12-25
implements rename command

### LINKED ISSUE #62: Rename project / tag
is there anything planed to rename a project / tag? At the moment I'm doing this with search&replace on the frames file directly (which is possible but most probably not indented ;-)).

I may also be able to contribute with a PR if we agree on a syntax.

    watson rename --project old_project new_project
    watson rename --tag old_tag new_tag

would be a first simple approach. At a later point, we could also integrate some sort of interactivity to ask on a per entry basis for renames.

Or am I heading in a wrong direction since there already is such a command or something?

awesome tool btw :+1:!

larsborn:
should fix #62. Since it is my first contribution, better take a close look at it ;-)

### DIFF
--- a/watson.completion
+++ b/watson.completion
@@ -75,6 +75,18 @@ _watson_complete () {
             COMPREPLY=($(compgen -W "$tags" -- ${cur}))
           fi
           ;;
+        rename)
+          case "$3" in
+            --project)
+              projects="$(watson projects)"
+              COMPREPLY=($(compgen -W "$projects" -- ${cur}))
+              ;;
+            --tag)
+              tags="$(watson tags)"
+              COMPREPLY=($(compgen -W "$tags" -- ${cur}))
+              ;;
+          esac
+          ;;
       esac
       ;;
   esac
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -971,3 +971,40 @@ def merge(watson, frames_with_conflict, force):
     watson.frames = original_frames
     watson.frames.changed = True
     watson.save()
+
+
+@cli.command()
+@click.option('--project', default=None, nargs=2)
+@click.option('--tag', default=None, nargs=2)
+@click.pass_obj
+def rename(watson, project, tag):
+    # input validation
+    if not tag and not project:
+        raise click.ClickException(style('error', 'You have to call rename with --project or --tag'))
+
+    if tag:
+        old_tag, new_tag = tag
+        if old_tag not in watson.tags:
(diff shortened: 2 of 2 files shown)


watson/cli.py
```diff
@@ -971,3 +971,40 @@ def merge(watson, frames_with_conflict, force):
...
+
+    if project:
+        old_project, new_project = project
+
+        if old_project not in watson.projects:
+            raise click.ClickException(style('error', 'Project "%s" does not exist' % old_project))
+
+    # rename project/tag
+    for frame in watson.frames:
+        watson.frames[frame.id] = (
```
k4nar: Could you use something like:
```python
if frame.project == old_project:
    frame._replace(frame=new_project)
```

That way if we change the format of the frame we will not have to edit this function.

k4nar:
Great, thank you for this contribution :+1: !

However, you have some minor PEP8 issues, could you fix them (it's just too long lines basically) ?

jmaupetit:
Thank you for this @larsborn !

watson/cli.py
```diff
@@ -971,3 +971,40 @@ def merge(watson, frames_with_conflict, force):
     watson.frames = original_frames
     watson.frames.changed = True
     watson.save()
+
+
+@cli.command()
+@click.option('--project', default=None, nargs=2)
```
jmaupetit: Short style flag `-p` would be relevant here

watson/cli.py
```diff
@@ -971,3 +971,40 @@ def merge(watson, frames_with_conflict, force):
...
+@click.option('--tag', default=None, nargs=2)
```
jmaupetit: Idem with `-T` for cross commands consistency

watson.completion
```diff
@@ -75,6 +75,18 @@ _watson_complete () {
             COMPREPLY=($(compgen -W "$tags" -- ${cur}))
           fi
           ;;
+        rename)
+          case "$3" in
+            --project)
```
jmaupetit: Idem with `-p`

watson.completion
```diff
@@ -75,6 +75,18 @@ _watson_complete () {
...
+              projects="$(watson projects)"
+              COMPREPLY=($(compgen -W "$projects" -- ${cur}))
+              ;;
+            --tag)
```
jmaupetit: and `-T`

jmaupetit:
Seems like flake8 does not validate your PR (too long lines). Can you fix it?

PR #71, opened 2016-01-21
Improve log command

### LINKED ISSUE #48: With hours > 9, log output doesn't line up
An example from me, forgetting to stop my timer last night:

        f9b1aa1  16:24 to 16:25     01m 13s               watson  [personal, floss]
        d59db15  16:25 to 07:23  14h 57m 45s               watson  [personal, floss]

I would have expected to see

        f9b1aa1  16:24 to 16:25      01m 13s               watson  [personal, floss]
        d59db15  16:25 to 07:23  14h 57m 45s               watson  [personal, floss]

Given that there are 168 hours in a week, even disallowing for overlap in theory someone *could* have 100 hours one task, and up to 68 hours distributed over the rest of their time.

Though I guess technically you could get the log for *more* than a week, so perhaps the correct thing to do here would be something like `len(str(days * 24))` to get the format width specifier (and I guess +1, for the 'h' field)

jmaupetit:
* Fix #48 (frame duration alignment when it spans more than 9h)
* Add daily total (pretty neat). Thank @themouette for the suggestion!

### DIFF
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -527,7 +527,7 @@ def log(watson, from_, to, projects, tags):
         lines.append(style('date', "{:dddd DD MMMM YYYY}".format(day)))
 
         lines.append('\n'.join(
-            '\t{id}  {start} to {stop}  {delta:>10}  {project}  {tags}'.format(
+            '\t{id}  {start} to {stop}  {delta:>11}  {project}  {tags}'.format(
                 delta=format_timedelta(frame.stop - frame.start),
                 project=style('project',
                               '{:>{}}'.format(frame.project, longest_project)),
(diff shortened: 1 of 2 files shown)


watson/cli.py
```diff
@@ -524,10 +524,21 @@ def log(watson, from_, to, projects, tags):
         frames = sorted(frames, key=operator.attrgetter('start'))
         longest_project = max(len(frame.project) for frame in frames)
 
-        lines.append(style('date', "{:dddd DD MMMM YYYY}".format(day)))
+        daily_total = reduce(
+            operator.add,
+            (frame.stop - frame.start for frame in frames)
+        )
```
k4nar: :+1:

k4nar:
LGTM :+1:

PR #76, opened 2016-01-26
Fix handling exception

dblN:
The codes inside `else` are executed if no exception throws in `try`. If I get it right, these lines are expected to catch any exception other than `IOError` and `ValueError`. So, I replace `else` with `except`.

### DIFF
--- a/watson/watson.py
+++ b/watson/watson.py
@@ -87,9 +87,11 @@ class Watson(object):
                 raise WatsonError(
                     "Invalid JSON file {}: {}".format(filename, e)
                 )
-        except:
+        except Exception as e:
             raise WatsonError(
-                "Impossible to open JSON file in {}".format(filename)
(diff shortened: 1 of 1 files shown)


k4nar:
Indeed, nice catch! However, I think we should have `except Exception as e` and print the exception in the message. Because if we get there, it's because we have some nasty error. I's not because the file is inaccessible or is not a valid JSON file, so it must be more like an alien attack or something :) .

dblN:
@k4nar ok. then, something like this?
```python
...
except Exception as e:
    raise WatsonError(
        "Unexpected error while loading JSON file {}: {}".format(filename, e)
    )
...
```

k4nar:
Yes exactly :) .

k4nar:
Thanks :+1: !

PR #80, opened 2016-02-06
[WIP] Add tests for the cli

jmaupetit:
Preliminary work on #5 that should fail due to [`WatsonError` monkey patching in `cli.py`](https://github.com/TailorDev/Watson/blob/cli-tests/watson/cli.py#L57). Still searching how to fix this smartly. Any help/advice would be appreciated.


willdurand:
:+1:

PR #81, opened 2016-02-08
Add support for WATSON_DIR in the fuzzer

jmaupetit:
The fuzzer script is potentially harmful, we needed to add support for the application directory overriding.


k4nar:
Nice idea. LGTM.

willdurand:
:+1:

PR #84, opened 2016-02-09
Fix watson directory

jmaupetit:
As suggested by @SpotlightKid


k4nar:
Oh yeah sorry about that, I'm always a bit confused about the behavior of `.get` and `.pop`.

jmaupetit:
My bad. I have added a test for this.

willdurand:
:+1:

PR #91, opened 2016-02-20
Require "pytest-runner" as development dependency

loonies:
Requiring "pytest-runner" as an install dependency adds unnecessary
bloat to the end user package. Instead, only development build should
require "pytest-runner" as a dependency.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,4 +9,5 @@ flake8
 mock<=1.0.1
 pytest
 pytest-datafiles
+pytest-runner
 tox
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
 arrow
 click
-pytest-runner
 requests


willdurand:
> I think you are right: we should remove dependencies for tests in a typical stand alone installation

:+1:

jmaupetit:
@k4nar ready to be merged?

willdurand:
Also, it would be nice to upgrade `mock`

loonies:
On an unrelated note, what's the usage of the `-r` option in the [requirements-dev.txt][1] file?

    -r requirements.txt

It is skipped in the [setup.py][2] file.

    if line.startswith(('#', '-r', '--')):
        continue

[1]: https://github.com/TailorDev/Watson/blob/00e22acaa416483dced8e9ce12fe312ee56e8f9e/requirements-dev.txt#L1
[2]: https://github.com/TailorDev/Watson/blob/00e22acaa416483dced8e9ce12fe312ee56e8f9e/setup.py#L28-L29

k4nar:
@loonies : it means "install this requirements file", so you can only do `pip install -r requirements-dev.txt` and it will install all the requirements.

@jmaupetit : It's good for me, I don't know why it wasn't already like that.

loonies:
Thanks, I've found more info in the _pip_ documentation.

k4nar:
@jmaupetit : Is there still any pending question or can we merge it ?

jmaupetit:
LGTM

PR #92, opened 2016-02-21
[WIP] Add zsh completion

gdetrez:


jmaupetit:
Thank you for this! Can you please update the documentation (see: http://tailordev.github.io/Watson/#command-line-completion)?

k4nar:
From [this wiki page](http://zshwiki.org/home/convert/bash) I understand that ZSH is able to use Bash completion functions. Wouldn't that be simpler (if possible of course)?

gdetrez:
@jmaupetit Sure. I'll try to do it this week

@k4nar I didn't know about this, thanks for the tips.

I tried it and the answer is *nja* (whitch is Swedish for "kind-of maybe but not really" :stuck_out_tongue_winking_eye:): It works, but it is a bit limited.

For instance, here's what I get when completing commands with the bash completion function:
selection_009
Compared to what the native completion does:
selection_007

Perhaps more useful with frame ids:
selection_010
selection_008

jmaupetit:
@gdetrez your implementation of Watson's completion makes me want to switch to zsh :smile_cat: I think we need to have comparable features/displays with bash completion.

molovo:
Really nice work @gdetrez, much simpler than the one I had written.

I am getting an issue though autocompleting frames within the `edit` and `remove` commands. First, I didn't have `jq` installed, so can we please make sure that this dependency is well documented, or find a way to display the list of frames without dates if `jq` isn't installed? Since installing `jq` though, I'm getting the following error:

```
_values:compvalues:10: not enough arguments
jq: error: Could not open file .config/watson/frames: No such file or directory
```

What is this file?

gdetrez:
@molovo thanks for the feedback

I noticed the missing '+' and fixed it but haven't updated the branch yet.

The path to the frame file should be absolute (also fixed in my local copy) but the dependency on `jq` might be a problem. It might be possible to make it optional: if you don't have `jq` it only shows you the tags (like the bash completion) but if you have it, you see the nice annotations as well.

gdetrez:
I'll try to push my changes when I get home tonight

gdetrez:
Sorry for the delay. I pushed my fixes and added the following  caching.

@molovo: Thanks for the idea, I didn't realize zsh had a caching mechanism built-in :heart:. Can you try again and check that it works now? It should also gracefully fall back on using `watson frames` if `jq` is not installed.

gdetrez:
@jmaupetit I have now updated the documentation.

BTW I'm not sure the completion system in bash can do that but in any case you should definitively give zsh a try :heart_eyes_cat:.

molovo:
Very close, projects and tags working very nicely, but I still can't get frames to autocomplete. What should I put in the frames file?

molovo:
Nevermind, I found the frames file. Frame completion is working really well, nice and responsive now you've added caching. Nice work.

Cheeky request, and might not be possible, but how easy would it be to sort the frames by date, and possibly remove the T and Z from the timestamp to improve readability.

gdetrez:
> Nevermind, I found the frames file. Frame completion is working really well, nice and responsive now you've added caching. Nice work.

@molovo Do you mean that the `frames` file is in a different location on your machine?

> Cheeky request, and might not be possible, but how easy would it be to sort the frames by date, and possibly remove the T and Z from the timestamp to improve readability.

I think jq has some sorting functions so I can take a look.

jmaupetit:
Thank you guys! Please ping us when it's ready for a final review.

molovo:
Yeah, on OSX it's at `~/Library/Application\ Support/watson/frames`. This is set by the `$WATSON_FOLDER` env variable, so might be worth checking for that? Once I set `WATSON_FOLDER=~/.config/watson` the completion worked fine.

jmaupetit:
@molovo The env var is `WATSON_DIR`, see: http://tailordev.github.io/Watson/user-guide/configuration/#application-folder

molovo:
@jmaupetit you are right, just finger trouble :)

jmaupetit:
For `zsh` n00bs like me, please add something like the following to the documentation:

```zsh
# Make sure that your .zshrc enables compinit
autoload -Uz compinit && compinit
```

jmaupetit:
Thx @gdetrez !

gdetrez:
I have now changed the way the frames are completed: I use the output of `watson log` instead of parsing the frame files with `jq`. This means that it now only completes the frames that are shown in the default log (one week here but might be configurable?). My guess is that this is what one would want after using watson for a while, but I might be wrong...  I think it is the most usable version so far.

I even made a screencast :laughing: 

[asciicast](https://asciinema.org/a/bhnzkuu6vl9560rgxcni8yhum)

The only problem that remains is getting the correct path to the frame file. I added support for the WATSON_DIR variable but that'll only work if the user has set it explicitely. @jmaupetit is there a portable way to get the default watson directory?

k4nar:
Wow that looks absolutly amazing! Great work!

> is there a portable way to get the default watson directory?

If `WATSON_DIR` is not defined, the path depends on the platform:
- MacOSX: ~/Library/Application Support/watson/config
- Windows: C:\Users\<user>\AppData\Local\watson\config
- Linux: ~/.config/watson/config

molovo:
:+1: :+1: :+1: That is an awesome enhancement @gdetrez. Really nice work

jmaupetit:
:+1: Great! @gdetrez you made me switch to zsh :heart:

gdetrez:
Thanks @k4nar. I added some more tests to guess the correct directory when on OSX. @molovo  can you try and see if it works?

Maybe it would make sens at some point if watson itself could output the directory, like for config values.

gdetrez:
@molovo thanks!

@jmaupetit  :tada:

watson.zsh-completion
```diff
@@ -0,0 +1,180 @@
...
+    'report:Display a report of the time spent on each...'
+    'restart:Restart monitoring time for a previously...'
+    'start:Start monitoring time for the given project.'
+    'status:Display when the current project was started...'
+    'stop:Stop monitoring time for the current project.'
+    'sync:Get the frames from the server and push the...'
+    'tags:Display the list of all the tags.'
+  )
+  _describe -t commands 'commands' commands "$@"
+}
```
k4nar: Is it possible to move this block in the top of the file, in order to ease the maintenance (when we add a command for instance) ?

molovo:
I'm getting some nasty completion options when trying to complete frames. Looks like the output of some usage text:

[asciicast](https://asciinema.org/a/0zdxir62a72fv7e5tyu4fgppd)

I'll keep looking into it.

watson.zsh-completion
gdetrez: Done

gdetrez:
@molovo That's strange. What's your zsh version?

molovo:
@gdetrez 5.2

watson.zsh-completion
```diff
@@ -0,0 +1,186 @@
...
+
+
+# ~~~ LOCATING THE FRAMES FILE ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+if [[ -n "$WATSON_DIR" ]]; then
+  _watson_frame_file=${WATSON_DIR}/frames
+elif [[ "$OSTYPE" =~ "^darwin" ]]; then
+  _watson_frame_file=${HOME}/Library/Application Support/frames
+else
+  _watson_frame_file=${HOME}/.config/watson/frames
+fi
```
k4nar: I really think this is the way to go. An alternative would be to get the output of `EDITOR=echo watson config -e` but that a ugly hack imo.

gdetrez: Ok, well I'll leave it like this then.

gdetrez:
@molovo 5.1.1 here. I tried 5.2 and it works too. Maybe a configuration difference? Can you try with a blank config?

molovo:
Tried that, with no .zshrc at all I still get the same thing. Might be a mac-specific issue, I'll keep playing.

molovo:
Any chance you could let me know the ZSH completion options you have enabled?

loonies:
Note to myself: Update PKGBUILD

molovo:
@gdetrez I've managed to get a version of the `_watson_frames` completion function working on OSX. See my version [here](https://github.com/molovo/Watson/commit/b44bf41b3f84c1c5e738fffc4858da7d0f9ad628#diff-c3c677cb425905ad04a0af27a68e96efR120). Should just be a drop-in replacement for the `_watson_frames` function in your file.

gdetrez:
I rebased on master and added your version of `_watson_frames` slightly modified to show the frame IDs. @molovo can you check that it still works in your zsh?

molovo:
Yep, looking good :)

docs/index.md
```diff
@@ -78,8 +78,23 @@ $ python setup.py install
 
 ### Command line completion
 
+#### Bash
+
 If you use a Bash-compatible shell, you can install the `watson.completion` file from the source distribution as `/etc/bash.completion.d/watson` - or wherever your distribution keeps the Bash completion configuration files. After you restart your shell, you can then just type `watson` on your command line and then hit `TAB` to see all available commands. Depending on your input, it completes `watson` commands, command options, projects, tags and frame IDs.
 
+#### ZSH
+
+If you use zsh, copy the file `watson.zsh-completion` somewhene in your
```
jmaupetit: typo: somewhe**R**e

gdetrez:
Sorry, I forgot to push the typo fix. @jmaupetit I think this is ready for the final review :smile_cat:.
(still using zsh? :wink:)

jmaupetit:
Yep, I am still in love with zsh. Thank you for this!

PR #96, opened 2016-02-23
Add basic ZSH completion

molovo:
Doesn't quite cover all bases yet, but it's a pretty good start.

Adds completion for:
  - All options and commands
  - Frames on the `edit` and `remove` commands
  - Projects and tags (with the preceding +) on the `start` command
  - Options on the `log` and `report` commands (including tag and project completion for those options)

Happy to add any other completions if you think I've missed anything important.

### DIFF
--- a/watson.zsh-completion
+++ b/watson.zsh-completion
@@ -1,0 +1,122 @@
+#compdef watson
+
+_watson() {
+  local curcontext="$curcontext" state line
+  local commands projects tags options
+  typeset -A opt_args
+
+  if ([[ ${+watson_commands} -eq 0 ]] || _cache_invalid watson_commands) \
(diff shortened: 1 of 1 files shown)


molovo:
To use the completion file, just rename `watson.zsh-completion` to `_watson` and place it somewhere in your `$fpath`.

jmaupetit:
Thank you for this @molovo! Have you seen @gdetrez PR #92? Can you join your forces on this? :heart:

molovo:
Ah sorry, I wrote this last week before that pull request was created, and didn't check before submitting. @gdetrez's implementation actually seems to be far more extensive than mine anyway, so I'm inclined to close this and help him out with his.

jmaupetit:
:+1: Thank you for this.

PR #102, opened 2016-02-29
Support JSON output in report command

timofurrer:
Use `--json` switch to change report output from console output to JSON.

### DIFF
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -326,6 +326,11 @@ You can limit the report to a project or a tag using the `--project` and
 `--tag` options. They can be specified several times each to add multiple
 projects or tags to the report.
 
+You can change to output format by specifing the format via `--format`.
+Available formats are:
+    - rich (console output with human-readable formatted text)
+    - json (console output in json)
+
 Example:
 
 
@@ -367,6 +372,31 @@ Example:
             [reactor   8h 35m 50s]
             [steering 10h 33m 37s]
             [wheels   10h 11m 35s]
+    
+    $ watson report --format json
+    {
+        "projects": [
+            {
+                "name": "watson",
+                "tags": [
+                    {
+                        "name": "export",
+                        "time": 530.0
+                    },
+                    {
+                        "name": "report",
+                        "time": 530.0
+                    }
+                ],
+                "time": 530.0
+            }
+        ],
+        "time": 530.0,
+        "timespan": {
+            "from": "2016-02-21T00:00:00-08:00",
+            "to": "2016-02-28T23:59:59.999999-08:00"
+        }
+    }
 
 ### Options
 
@@ -381,6 +411,7 @@ Flag | Help
 `-d, --day` | Reports activity for the current day.
 `-p, --project TEXT` | Reports activity only for the given project. You can add other projects by using this option several times.
 `-T, --tag TEXT` | Reports activity only for frames containing the given tag. You can add several tags by using this option multiple times
+`-f, --format [rich|json]` | Output format for the report
 `--help` | Show this message and exit.
 
 ## `restart`
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -326,10 +326,8 @@ You can limit the report to a project or a tag using the `--project` and
 `--tag` options. They can be specified several times each to add multiple
 projects or tags to the report.
 
-You can change to output format by specifing the format via `--format`.
-Available formats are:
-    - rich (console output with human-readable formatted text)
-    - json (console output in json)
+You can change the output format for the report from *plain text* to *JSON*
+by using the `--json` option.
 
 Example:
 
@@ -411,7 +409,7 @@ Flag | Help
 `-d, --day` | Reports activity for the current day.
 `-p, --project TEXT` | Reports activity only for the given project. You can add other projects by using this option several times.
 `-T, --tag TEXT` | Reports activity only for frames containing the given tag. You can add several tags by using this option multiple times
-`-f, --format [rich|json]` | Output format for the report
+`-j, --json` | Format the report in JSON instead of plain text
 `--help` | Show this message and exit.
 
 ## `restart`
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -325,8 +325,12 @@ _SHORTCUT_OPTIONS = ['year', 'month', 'week', 'day']
               help="Reports activity only for frames containing the given "
               "tag. You can add several tags by using this option multiple "
               "times")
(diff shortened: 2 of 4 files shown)


jmaupetit:
Thank you @timofurrer ! JSON output seems a relevant feature. Can you share a concrete use case with us? I mean to which program do you pipe this output to?

Side question: did you update `commands.md` with `make docs`?

timofurrer:
I don't have a specific tool at the moment. However, I plan to write an adapter for some company specific time tracking tools ;) .. and beside that I thought others might have the same idea so I just implemented it .

Yeah, I documented it -> See changeset. Is something wrong about it?

jmaupetit:
> I plan to write an adapter for some company specific time tracking tools

I am glad that you prefer using Watson than the time tracker of your company :heart: 

> Yeah, I documented it -> See changeset. Is something wrong about it?

Yep, saw it. That is why I asked. Nothing wrong, just noise.

docs/user-guide/commands.md
```diff
@@ -341,6 +366,7 @@ Flag | Help
 `-t, --to DATE` | The date at which the report should stop (inclusive). Defaults to tomorrow.
 `-p, --project TEXT` | Reports activity only for the given project. You can add other projects by using this option several times.
 `-T, --tag TEXT` | Reports activity only for frames containing the given tag. You can add several tags by using this option multiple times
+`-j, --json` | Outputs the report as JSON instead of writing to the console
```
SpotlightKid: JSON output still goes to the console, doesn't it? Just the format changes from formatted text with escape sequences to JSON. How can this be reworded?

watson/cli.py
```diff
@@ -339,6 +341,9 @@ def report(watson, from_, to, projects, tags):
     `--tag` options. They can be specified several times each to add multiple
     projects or tags to the report.
 
+    You can change to output format from the console output to JSON with
+    specifying the `--json` argument.
```
SpotlightKid: Again, the difference is not console vs something else but formatted text vs JSON.

watson/cli.py
```diff
@@ -322,8 +322,10 @@ def status(watson):
               help="Reports activity only for frames containing the given "
               "tag. You can add several tags by using this option multiple "
               "times")
+@click.option('-j', '--json', 'json_output', is_flag=True,
+              help="Outputs the report as JSON instead of writing to the console")
```
SpotlightKid: Again, output still goes to console (stdout), only format changes.
 
Also this line's length is currently >79 and make the flake8 test fail.

SpotlightKid:
@timofurrer Do you still plan to update this PR according to the feedback?

timofurrer:
Completely forgot about this PR. I'll have a look at it asap.

docs/user-guide/commands.md
```diff
@@ -296,43 +296,68 @@ Example:
...
-    
+
     apollo11 - 13h 22m 20s
             [brakes    7h 53m 18s]
             [module    7h 41m 41s]
             [reactor   8h 35m 50s]
             [steering 10h 33m 37s]
             [wheels   10h 11m 35s]
 
+    $ watson report --format json
```
yloiseau: Here you document `--format json` while the actual option is `--json`.
BTW, I think the `--format` is better since it leaves room to future other formats without adding lots of distinct options.

timofurrer:
It's definitely been a while - I apologize for that!

I've updated the PR and reworded the option description.
I've used `--format [rich|json]` instead of `--json`. I'm not quiet happy with the term `rich` though. Any suggestions?

SpotlightKid:
> I'm not quiet happy with the term rich though. Any suggestions?

Umm, not sure, how about `formatted` or `plaintext` or `plain`?

willdurand:
I would keep `--json` as it is short and useful. There is no need for other formats yet, so we can keep things simple. And it does not mean `--format` will not be implemented in the future.

timofurrer:
> Umm, not sure, how about formatted or plaintext or plain?

Well, the json output is `formatted`, too. And it's not really plaintext because of the colors.

> I would keep --json as it is short and useful. There is no need for other formats yet, so we can keep things simple. And it does not mean --format will not be implemented in the future.

What about a shortcut to `--format json`? I agree, it does not make a lot of sense to expose an API (`--format`) which is *not really used* (well, at this point it is used for some extend, but since we only have to choices - we could consider it a flag)

I can do both. Who makes the calls here?

jmaupetit:
Thanks for the update @timofurrer ! I would also keep the `--json` flag and we will add a shortcut later when the `--format` api will make sense.

timofurrer:
@jmaupetit @willdurand I've updated the PR according to your suggestion to use `--json` instead of `--format`.

watson/cli.py
```diff
@@ -325,8 +325,11 @@ def status(watson):
               help="Reports activity only for frames containing the given "
               "tag. You can add several tags by using this option multiple "
               "times")
+@click.option('-j', '--json', 'format_json', is_flag=True,
+              help='Format the report in JSON instead of plaintext')
```
jmaupetit: `s/plaintext/plain text/`

watson/cli.py
```diff
@@ -346,6 +349,9 @@ def report(watson, current, from_, to, projects, tags, year, month, week, day):
     `--tag` options. They can be specified several times each to add multiple
     projects or tags to the report.
 
+    You can change the output format for the report from *plaintext* to *JSON*
```
jmaupetit: Idem: `s/plaintext/plain text/`

timofurrer:
@jmaupetit okay. 

I'm not quiet sure what you mean with *translatable strings* - I couldn't find a consistent pattern of the use for `"` and `'` ...

But I guessed and pushed - Have a look at it :beers:

jmaupetit:
Thank you for these changes. I am sorry to insist, but there are some double quotes left in `cli.py` (l. 430-467). Once fixed, I'll merge this PR. 💪

timofurrer:
Done.

I think some clear guide lines would really help people to contribute efficiently ... ;)

willdurand:
> I think some clear guide lines would really help people to contribute more efficiently ... ;)

I do agree with you, especially here since I do not know what _translatable strings_ are either. Maybe @jmaupetit could give an explanation here, and we will update the `CONTRIBUTING` file/doc accordingly.

jmaupetit:
@timofurrer I am sorry if you felt frustrated after submitting code to this PR. This coding convention is implicit, but required to maintain consistency over the project. We need to work on this.

To briefly explain it, _translatable strings_ are literal strings containing natural language messages that may be translated during the internationalization of the app. All other strings acting as "identifiers" are single quoted (dictionary keys, etc.)

PR #103, opened 2016-02-29
Specify correct source directory for flake8 and pytest in tox.ini

timofurrer:
If the tests directory is not specified this causes problems if the
developer has virtualenvs setup in the Watson root directory.
The packages in there could include tests which are then run by tox.

Same for flake8 and the watson and tests directory.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,9 +5,9 @@ envlist = flake8,py27,py33,py34,py35
 deps=pytest
      mock
      pytest-datafiles
-commands=py.test -vs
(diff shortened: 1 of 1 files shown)


jmaupetit:
:+1:

k4nar:
:+1: 

We should add `scripts/` to the flake8 command though.

timofurrer:
I've added `scripts/` to the `flake8` call.

k4nar:
Thanks!

PR #104, opened 2016-02-29
Suggest to run tox instead of pytest in hack docs

timofurrer:


jmaupetit:
Too long and too greedy, I think Travis is here to help with that.

k4nar:
Yes, plus tox only works if you have all the versions of Python supported by Watson installed in your environment.

Instead of replacing pytest by tox in the docs, maybe we could add something about it.

jmaupetit:
:+1: to add something about `tox` vs `pytest` in the docs.

SpotlightKid:
I think contributors can be expected to go to the trouble of at least installing tox and python 2.7 and 3.5 (or  python 3.4, if their distro only provides that). Just makes it easier for them to avoid surprises from failing Travis jobs.

We could remove `py33` and `py34` envs from `tox.ini`, so they don't run by default but still run them on Travis (via `.tavis.yml`).

SpotlightKid:
Tox >= 1.7.2 has a config option [skip_missing_interpreters](https://testrun.org/tox/latest/config.html#confval-skip_missing_interpreters=BOOL), which, when set to `True`, allows to run the tests via tox even if you don't have the Python versions for all specified tox environments installed. If we add this to `tox.ini`, I think we can safely recommend to run `tox` instead of `py.test`.

SpotlightKid:
Closing this in favour of #133.

PR #106, opened 2016-02-29
Remove pytest-runner from requirements.txt

timofurrer:


jmaupetit:
Already addressed in: https://github.com/TailorDev/Watson/pull/91

PR #107, opened 2016-03-05
Leverage arrow timedelta property to simplify format_timedelta()

firecat53:
Hey, I was perusing the code looking to make some feature requests and I noticed completely by accident that the format_timedelta function can be simplified by using a feature of arrow that returns `hh:mm:ss` on a `str(delta)`. Nice! Works for any value from 0 on up. If it goes into days, it just looks like `x days hh:mm:ss` which still works perfectly with this PR.

Thanks for a great app!

Scott

### DIFF
--- a/watson/utils.py
+++ b/watson/utils.py
@@ -40,9 +40,20 @@ def style(name, element):
 def format_timedelta(delta):
     """
     Return a string roughly representing a timedelta.
+
+    For reference, str(timedelta) returns [x days,] hh:mm:ss
+
     """
     neg = '-' if int(delta.total_seconds()) < 0 else ''
-    res = "{}h {}m {}s".format(*str(abs(delta)).split(":"))
+    fmt = ["{}h ", "{}m ", "{}s"]
+    td = str(abs(delta)).split(":")
+    td[:0] = td.pop(0).split(',')
+    if len(td) == 4:
+        add_hours = int(td.pop(0).split()[0].strip()) * 24
+        td[0] = str(int(td[0]) + add_hours)
+    res = str()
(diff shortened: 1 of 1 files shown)


watson/utils.py
```diff
@@ -41,25 +41,9 @@ def format_timedelta(delta):
...
-    if total >= 60:
-        mins = seconds // 60
-        stems.append('{:02}m'.format(mins))
-        seconds -= mins * 60
-
-    stems.append('{:02}s'.format(seconds))
-
-    return ('-' if neg else '') + ' '.join(stems)
+    neg = '-' if int(delta.total_seconds()) < 0 else ''
+    res = "{}h {}m {}s".format(*str(abs(delta)).split(":"))
```
jmaupetit: Shouldn't it be?

```python
res = "{}h {:02}m {:02}s".format(*str(abs(delta)).split(":"))
```

jmaupetit:
:+1: Thank you for this. It seems pretty neat!

watson/utils.py
firecat53: The python datetime module actually takes care of the formatting for us.
See the str() function docs:
https://docs.python.org/3/library/datetime.html#timedelta-objects.

Thanks!
Scott

firecat53:
Realized the initial PR didn't handle times > 24 hours. Fixed that, plus hide any values that are zero. It's a little different than you have currently, but I think it works and is readable.

Thanks!
Scott

firecat53:
I'm going to close this PR and open a new one with #107 and #108 merged together. Sorry for the noise!

PR #108, opened 2016-03-05
Rounding of work times up to nearest x minutes

firecat53:
This is an initial commit (no docs or tests yet) of a feature to offer rounding up of work times. I wanted to make sure this would be of interest before I finalize it.

Rationale: I don't think most people track their work times to the nearest second. Also, a number of contractors round times up to the nearest hour or half-hour. 

Changes: added a command-line switch --round-to (-r). Defaults to zero (no rounding). The rounding feature rounds the totals for each day based on the start time. So if you have multiple frames in a day, it will sum all of the frames, then present the rounded-up total for that day. It works with both the 'report' and 'log' options. The time per day is also summed and rounded for each tag in the log view.

I've attached a file with sample outputs with and without rounding.

Let me know if this makes sense and seems useful. 

Scott
[log.txt](https://github.com/TailorDev/Watson/files/160088/log.txt)

PR #109, opened 2016-03-11
Add rounding option for total work times and changes default report view style.

firecat53:
This PR combines #107 and #108. To summarize:

- Adds ability to round up total work times  in the log and report views. Total work time for each project for each calendar day is summed and rounded up each day, then those values are added for the total time during the report/log period.
- Adds a 'By Date' summary to the report view in addition to the original 'By Tag' view
  - Note: the 'By Tag' view is rounded to the nearest minute regardless of desired rounding. This is due to the rounding differences when rounding all work for one day vs rounding up just the work for one tag during a day. The sum of the 'By Date' hours will be equal to the total value at the top of the report.
  - I added the 'By Date' view to the report because it seems useful to me as a contractor, and because I wanted the sum of the hours to match the total at the top. It made more sense to me to round up total work for the entire day on all tags rather than rounding up each tag each day. I know...it's confusing to read at first!!
- Adds test for format_timedelta and round_time methods
- Changes default report and log rounding to 1 minute (seconds only displayed for each frame in the log view).
- Adds configuration file option for default rounding value

Sorry for the wall of text. Hopefully trying it out and testing the different rounding options will explain it better than my words :smile:

Example report output:

```
$ watson report -f 2016-03-01 --round_to 15
Tue 01 March 2016 -> Thu 10 March 2016

projectname - 5h 30m

        By Tags (unrounded) 
        -------------------
        [tag1         2h 03m]
        [tag2         1h 03m]
        [tag3         1h 28m]
        [tag4         0h 37m]

        By Date
        -------
        [2016-03-04       1h 45m]
        [2016-03-07       0h 30m]
        [2016-03-08       1h 30m]
        [2016-03-10       1h 45m]
```

TODO:

- Add tests for total_by_day and total_by_each_day
- Edit the help text for reports to reflect the new output style
- Squash the commits before merging

### DIFF
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -156,6 +156,19 @@ Example:
     Wednesday 16 April 2014 (5h 19m 18s)
             02cb269  09:53 to 12:43   2h 50m 07s  apollo11  [wheels]
             1070ddb  13:48 to 16:17   2h 29m 11s  voyager1  [antenna, sensors]
+    
+    $ watson log --from 2016-02-16 --to 2016-03-15 --round_to 15
+    Friday 04 March 2016 (1h 45m 00s)
+            d7f7cd0  13:57 to 14:17      19m 23s  apollo11  [wheels]
+            4cc9841  14:27 to 15:11      43m 42s  apollo11  [lens]
+            ce3d2aa  16:29 to 17:06      36m 41s  apollo11  [reactor]
+    
+    Tuesday 23 February 2016 (2h 30m 00s)
+            6931d90  12:04 to 12:06      02m 02s  apollo11  [antenna]
+            f0d473c  12:06 to 13:15   1h 08m 32s  apollo11  [antenna]
+            678d9af  13:55 to 14:09      14m 04s  apollo11  [antenna]
+            ae51e4d  14:28 to 14:51      23m 18s  apollo11  [antenna]
+            6fe0aa3  20:26 to 21:01      34m 11s  apollo11  [antenna]
 
 ### Options
 
@@ -165,7 +178,7 @@ Flag | Help
 `-t, --to DATE` | The date at which the log should stop (inclusive). Defaults to tomorrow.
 `-p, --project TEXT` | Logs activity only for the given project. You can add other projects by using this option several times.
 `-T, --tag TEXT` | Logs activity only for frames containing the given tag. You can add several tags by using this option multiple times
-`-r, --round-to INTEGER` | Rounds log values up to the nearest x minutes.Defaults to no rounding.
+`-r, --round-to INTEGER` | Rounds daily totals up to the nearest x minutes. Defaults to no rounding.
 `--help` | Show this message and exit.
 
 ## `merge`
@@ -340,6 +353,16 @@ Example:
             [reactor   8h 35m 50s]
             [steering 10h 33m 37s]
             [wheels   10h 11m 35s]
+    
+    $ watson report --from 2016-02-10 --to 2016-03-30 --round_to 15
+    Wed 10 February 2016 -> Wed 30 March 2016
+    
+    apollo11 - 22h 45m 00s
+            [brakes   1h 00m 00s]
+            [module  13h 00m 00s]
+            [reactor  7h 15m 00s]
+            [steering 1h 30m 00s]
+            [wheels      45m 00s]
 
 ### Options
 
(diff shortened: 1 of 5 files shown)


loonies:
Diff has mixed `--round-to` and `--round_to`. I would prefer `--round-to` instead of `--round_to` as a CLI option.

jmaupetit:
Thank you again @firecat53 for your work on this! Just to share my thoughts on this PR:

1. Changes you made to the `report` command output should be treated separately in a new PR as it's not related to the new `--round-to` option you propose here.
2. I would love to have something more flexible for rounding in working days for example.

WDYT?

firecat53:
@jmaupetit -- 

1. Well, I _can_ separate the two. However, the reason I ended up combining the two aspects of the PR is that with the existing report format there's no easy way to get the total of the per-tag times to match the rounded report total. For example:

    Tue 01 March 2016 -> Thu 17 March 2016

    project - 11h 45m 
            [motor          8h 30m ]
            [suspension 1h 15m ]
            [phone         1h 30m ]
            [wiring               45m ]

You can see the project total for the period differs from the sum of the individual tags because the total is rounded by total work done each day, not by work per each tag per day. It's confusing, but the math works out if you try a few examples.

SO, if that's ok, I can separate the two PRs.

2. Can you give me some examples of what you had in mind for additional rounding options? Not sure what you meant by 'rounding in working days' ?

Thanks!
Scott

jmaupetit:
>  there's no easy way to get the total of the per-tag times to match the rounded report total

I think only the total should be rounded in this PR.

> Not sure what you meant by 'rounding in working days' ?

It would be nice to be able to finely-tune the rounding unit: minute/hours is a really good start, but what if I decide that the rounding unit should be a 7h slot (a working day) or 35h (a working week)?

firecat53:
> It would be nice to be able to finely-tune the rounding unit: minute/hours is a really good start, but what if I decide that the rounding unit should be a 7h slot (a working day) or 35h (a working week)?

I'm still not sure what you mean...just using hours should get the effect you want, right? If you work two 5 hour days and want to round to the nearest 7 hour day, then just `-r 420` and each of those 5 hour days will be rounded to 7 hours, giving a total of 14 hours for those two days.

Am I missing something? :smile:

Scott

jmaupetit:
You are right, but we will still need to convert rounding to "working days", and, this is not what we expect from such command.

To be more explicit, I would love something like:

```bash
$ watson report -f 2016-03-01 --round-to 1 --round-to-unit working_day --round-to-strategy up
Tue 01 March 2016 -> Thu 17 March 2016

project - 2 working days (12h 05m)
        [motor      8h 34m ]
        [suspension 1h 24m ]
        [phone      1h 18m ]
        [wiring        49m ]
```

As you can see from the `report` command output, we have 3 different options for rounding (see explanation below). On a per-project basis, the report contains:

* rounding in specified unit (with plural support)
* total exact time in parenthesis
* per-tag details in exact time

You can define your own units (in minutes) in the configuration file, with something like:

```ini
[options]
# those are defaults overridable via command options
round_to = 1  # defaults to 1
round_to_unit = working_day  # defaults to minute
round_to_stategy = nearest  # possible values are up, down and nearest (default)

[units]
working_day = 420  # 7 * 60 minutes
working_week = 2100  # 5 * 7 * 60 minutes
```

Being able to specify the rounding strategy seems critical to me, as working rules depend on this. WDYT?

Again, thank you for you work! Please let us know if you need more help on this.

SpotlightKid:
As reporting and logging commands and options get more sophisticated and complex, it might better to split them out into a separate program (e.g. `watson-report`). Calling it could still be integrated into the main `watson` command (http://click.pocoo.org/5/complex/ might be relevant here), though I don't particularly like programs with too many/complex sub-commands (`git` being an especially appalling example).

tremby:
I was just taking a look at Watson to see if it would suit me, and couldn't see this feature in the docs. It's make-or-break for me, so I'm hoping this gets finished and merged. But I think I have more requirements which this may not entirely fill:

- Ability to round (closest) or round up or to round down to unit size
- Ability to have a hard minimum (i.e. if rounding unit is 15 minutes, and strategy is "nearest", 2 minutes on a project still counts as 1 unit of 15 minutes, but 17 minutes also counts as 15 minutes)
- Ability to choose whether the minimum and rounding take place per frame, or per day, or per report
- Ability to have different options of these kinds applied to different projects

Are there plans to support any of these features?

PR #111, opened 2016-03-17
Added basic dmenu script to start/stop and view status

firecat53:
Another addition I found useful for starting/stopping Watson when I don't have a terminal open. Just add a keybinding :)

Let me know what you think!

Scott

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -80,6 +80,10 @@ $ python setup.py install
 
 If you use a Bash-compatible shell, you can install the `watson.completion` file from the source distribution as `/etc/bash.completion.d/watson` - or wherever your distribution keeps the Bash completion configuration files. After you restart your shell, you can then just type `watson` on your command line and then hit `TAB` to see all available commands. Depending on your input, it completes `watson` commands, command options, projects, tags and frame IDs.
 
+### Dmenu
+
+A dmenu script to start, stop and view status is available in the scripts folder. Copy or symlink the script to your bin folder and edit the configuration file if desired to add dmenu options (see the configuration documentation)
+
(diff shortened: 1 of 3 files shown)


jmaupetit:
Hi @firecat53! Thank you for this. It looks really nice, but I think it's a bit too specific to be integrated in the main repository. My advice would be to maintain it in a separate repository you own.

As side contributions are growing, we should document them in the documentation.

But. This is just my opinion. I would like community feedback on this.

k4nar:
Yeah, I think too that is should not be in the main repository. But it looks very great though.

SpotlightKid:
Yeah, I agree that contributions are better kept in separate repositories and linked in the docs.

BTW, I was also thinking about writing a little systray app to monitor and start and stop Watson. Not sure when I'll get around to it, so if anybody else wants to do it, don't hold back ;)

jmaupetit:
@SpotlightKid a few weeks ago I worked on an [electron](http://electron.atom.io/)-based systray to monitor `watson` status and be able to get (native) system notifications. I have not publicly released it yet, but I will soon :) Here is a preview:

screenshot_14_02_16_00_41

screenshot_14_02_16_00_29

SpotlightKid:
@jmaupetit Cool, that's even better than something GTK or Qt based.

firecat53:
Ok, thanks. Repo setup here:  https://github.com/firecat53/watson-dmenu

Let me know if anyone finds it useful or if there are additional features that could be added.

Scott

jmaupetit:
Thank you again for your work Scott. Let's keep in touch on this.

PR #113, opened 2016-04-12
[WIP] Add defaults tags for projects

jbarnoud:
Some projects are always used with the same tags. For instance, my
'python101' project should always have the 'teaching' tag attached to
it.

This commit adds a way to declare such default tags in the configuration
file, and have them applied. Default tags are expected to be declared in
a '[default_tags]' section:

    [default_tags]
    python101 = teaching
    watson = contrib python

With the sample configuration above, the 'teaching' tag will always be
attached to the 'python101' project, and the tags 'contrib' and 'python'
will always be attached to the project 'watson'.

Projects that do not appear in the '[default_tags]' section of the
configuration will not be affected by the change.

Note that the tags added by the new mechanism do not show in the output
of `watson start`. However, they appear in the output of `watson
status`, `watson log`, and of course `watson report`.

This pull request is still a work in progress. It still requires at least:
* [x] to update the documentation
* [x] to be properly tested

### DIFF
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -149,7 +149,7 @@ same tags are always attached to a particular project.
 
 These automatically attached tags are defined in the `[default_tags]` section
 of the configuration. Each option bellow that section is a project to which
-tags should be attached.
+tags should be attached. The entries should follow the pattern: `project = tag1 tag2`.
 
 You can set default tags for a project from the command line:
 
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -154,7 +154,7 @@ tags should be attached. The entries should follow the pattern: `project = tag1 
 You can set default tags for a project from the command line:
 
 ```
-$ watson watson config default_tags.python101 'teaching python'
+$ watson config default_tags.python101 'teaching python'
 ```
 
 This corresponds to the following configuration file snippets:
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -148,7 +148,7 @@ Tags can be automatically added for some projects. This is convenient when the
 same tags are always attached to a particular project.
 
 These automatically attached tags are defined in the `[default_tags]` section
-of the configuration. Each option bellow that section is a project to which
+of the configuration. Each option in that section is a project to which
 tags should be attached. The entries should follow the pattern: `project = tag1 tag2`.
 
 You can set default tags for a project from the command line:
@@ -175,7 +175,7 @@ $ watson start python101 +lecture
 Starting project python101 [lecture, teaching, python] at 19:28
 ```
 
-Default tags can contain space characters when wrote in between quotes:
+Default tags can contain space characters when written in between quotes:
 
 ```
 $ watson config default_tags.voyager2 'nasa "space mission"'
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -144,8 +144,8 @@ Globally configure how `time` should be formatted. All [python's `strftime` dire
 
 ### Default tags
 
-Tags can be automatically added for some projects. This is convenient when the
-same tags are always attached to a particular project.
+Tags can be automatically added for selected projects. This is convenient when
+the same tags are always attached to a particular project.
 
 These automatically attached tags are defined in the `[default_tags]` section
 of the configuration. Each option in that section is a project to which
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -142,6 +142,52 @@ Globally configure how `dates` should be formatted. All [python's `strftime` dir
 
 Globally configure how `time` should be formatted. All [python's `strftime` directives](http://strftime.org) are supported.
 
+### Default tags
+
+Tags can be automatically added for some projects. This is convenient when the
+same tags are always attached to a project.
+
+These automatically attached tags are defined in the `[default_tags]` section
+of the configuration. Each option bellow that section is a project to which
+tags should be attached.
+
+You can set default tags for a project from the command line:
+
+```
+$ watson watson config default_tags.python101 'teaching python'
+```
+
+This corresponds to the following configuration file snippets:
+
+```ini
+[default_tags]
+python101 = teaching python
+```
+
+With these default tags set, the tags "teaching" and "python" will
+automatically be attached to the project "python101":
+
+```
+$ watson start python101
+Starting project python101 [teaching, python] at 19:27
+
+$ watson start python101 +lecture
+Starting project python101 [lecture, teaching, python] at 19:28
+```
+
+Default tags can contain space characters when wrote in between quotes:
+
+```
+$ watson watson config default_tags.voyager2 'nasa "space mission"'
+```
+
+Or in the configuration file:
+
+```ini
+[default_tags]
+voyager2 = nana 'space mission'
+```
+
 ## Sample configuration file
 
 A basic configuration file looks like the following:
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -145,7 +145,7 @@ Globally configure how `time` should be formatted. All [python's `strftime` dire
 ### Default tags
 
 Tags can be automatically added for some projects. This is convenient when the
-same tags are always attached to a project.
+same tags are always attached to a particular project.
 
 These automatically attached tags are defined in the `[default_tags]` section
 of the configuration. Each option bellow that section is a project to which
@@ -178,14 +178,14 @@ Starting project python101 [lecture, teaching, python] at 19:28
 Default tags can contain space characters when wrote in between quotes:
 
 ```
-$ watson watson config default_tags.voyager2 'nasa "space mission"'
+$ watson config default_tags.voyager2 'nasa "space mission"'
 ```
 
 Or in the configuration file:
 
 ```ini
 [default_tags]
-voyager2 = nana 'space mission'
+voyager2 = nasa 'space mission'
 ```
 
 ## Sample configuration file
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -466,7 +466,7 @@ my project = A B
         assert watson.current['tags'] == ['A', 'B']
 
 
-def test_start_default_tags_with_tags(watson):
+def test_start_default_tags_with_supplementary_input_tags(watson):
     content = u"""
 [default_tags]
 my project = A B
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -407,6 +407,21 @@ def test_set_config(watson):
     watson.config.get('foo', 'bar') == 'lol'
 
 
+def test_get_default_tags(watson):
+    content = u"""
+[default_tags]
+single = tag1
+multiple = tag1 tag2 tag3
+spaces = tag1 'with space'
+    """
+
+    with mock.patch.object(ConfigParser, 'read', mock_read(content)):
+        assert watson._default_tags('single') == ['tag1']
+        assert watson._default_tags('multiple') == ['tag1', 'tag2', 'tag3']
+        assert watson._default_tags('spaces') == ['tag1', 'with space']
+        assert watson._default_tags('non defined') == []
+
+
 # start
 
 def test_start_new_project(watson):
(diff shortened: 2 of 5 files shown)


SpotlightKid:
Please note that tags can contain spaces. You should probably either change the format of values in the `default_tags` section to be comma-separated lists, or use the `split` function from stdlib `shlex` module to parse option values into tag lists.

jbarnoud:
@SpotlightKid I did not notice you could have spaces in tags. I'll see how I can solve this.

jbarnoud:
I added the doc and the tests, so the PR should be ready for review. Should I rebase now to squash my commits now, or after the review?

docs/user-guide/configuration.md
```diff
@@ -130,6 +130,52 @@ Globally configure how `dates` should be formatted. All [python's `strftime` dir
...
+
+```
+$ watson watson config default_tags.voyager2 'nasa "space mission"'
+```
+
+Or in the configuration file:
+
+```ini
+[default_tags]
+voyager2 = nana 'space mission'
```
jmaupetit: `s/nana/nasa`

docs/user-guide/configuration.md
```diff
@@ -130,6 +130,52 @@ Globally configure how `dates` should be formatted. All [python's `strftime` dir
...
+same tags are always attached to a project.
+
+These automatically attached tags are defined in the `[default_tags]` section
+of the configuration. Each option bellow that section is a project to which
+tags should be attached.
+
+You can set default tags for a project from the command line:
+
+```
+$ watson watson config default_tags.python101 'teaching python'
```
jmaupetit: I see `watson` twins here :smile_cat:

docs/user-guide/configuration.md
```diff
@@ -130,6 +130,52 @@ Globally configure how `dates` should be formatted. All [python's `strftime` dir
 
 Globally configure how `time` should be formatted. All [python's `strftime` directives](http://strftime.org) are supported.
 
+### Default tags
+
+Tags can be automatically added for some projects. This is convenient when the
+same tags are always attached to a project.
```
jmaupetit: to a **particular** project.

docs/user-guide/configuration.md
```diff
@@ -130,6 +130,52 @@ Globally configure how `dates` should be formatted. All [python's `strftime` dir
...
+
+These automatically attached tags are defined in the `[default_tags]` section
+of the configuration. Each option bellow that section is a project to which
+tags should be attached.
```
jmaupetit: I would have written something like:

Default tags are defined in the `[default_tags]` section. Each entry should follow the pattern: `project = tag1 tag2`.

tests/test_watson.py
```diff
@@ -361,6 +376,28 @@ def test_start_two_projects(watson):
...
+[default_tags]
+my project = A B
+    """
+
+    with mock.patch.object(ConfigParser, 'read', mock_read(content)):
+        watson.start('my project')
+        assert watson.current['tags'] == ['A', 'B']
+
+
+def test_start_default_tags_with_tags(watson):
```
jmaupetit: `test_start_default_tags_with_tags` -> `test_start_default_tags_with_supplementary_input_tags`

WDYT?

jbarnoud: Isn't that a bit long? I worry it would get truncated in the test output, making it less informative at the end.

watson/watson.py
```diff
@@ -425,3 +433,11 @@ def merge_report(self, frames_with_conflict):
                 merging.append(conflict_frame)
 
         return conflicting, merging
+
+    def _default_tags(self, project):
+        default_tags_raw = self.config.get('default_tags', project)
+        if default_tags_raw is None:
+            default_tags = []
+        else:
+            default_tags = shlex.split(default_tags_raw)
```
jmaupetit: I would prefer something like:

```python
default_tags = []
if default_tags_raw is not None:
    default_tags = shlex.split(default_tags_raw)
```

I'm trying to respect Object Calisthenics, see [rule 2](http://williamdurand.fr/2013/06/03/object-calisthenics/#2-don-t-use-the-else-keyword). Poke @willdurand :stuck_out_tongue_winking_eye:

watson/watson.py
```diff
@@ -247,6 +248,13 @@ def start(self, project, tags=None):
                 )
             )
 
+        default_tags = self._default_tags(project)
+        if default_tags and not restart:
+            if tags is None:
+                tags = default_tags
+            else:
+                tags = tags + default_tags
```
jmaupetit:
```python
tags += default_tags
```

jbarnoud: This would mutate the argument, do we want that side effect?

tests/test_watson.py
jmaupetit: I think it's more explicit. No worries, it won't be truncated in a standard terminal.

watson/watson.py
jmaupetit: Good question. @k4nar WDYT?

k4nar: I didn't have time to review the whole PR yet, sorry.

On this one I think that `tags = (tags or []) + default_tags` would be a good option.

docs/user-guide/configuration.md
```diff
@@ -130,6 +130,52 @@ Globally configure how `dates` should be formatted. All [python's `strftime` dir
 
 Globally configure how `time` should be formatted. All [python's `strftime` directives](http://strftime.org) are supported.
 
+### Default tags
+
+Tags can be automatically added for some projects. This is convenient when the
+same tags are always attached to a project.
+
+These automatically attached tags are defined in the `[default_tags]` section
+of the configuration. Each option bellow that section is a project to which
```
SpotlightKid: s/bellow/below/

But "Each option in this section" is better, IMHO.

docs/user-guide/configuration.md
```diff
@@ -130,6 +130,52 @@ Globally configure how `dates` should be formatted. All [python's `strftime` dir
...
+
+```
+$ watson start python101
+Starting project python101 [teaching, python] at 19:27
+
+$ watson start python101 +lecture
+Starting project python101 [lecture, teaching, python] at 19:28
+```
+
+Default tags can contain space characters when wrote in between quotes:
```
SpotlightKid: s/wrote/written/

watson/watson.py
SpotlightKid: This doesn't take into account when a default tag is also given on the command line. Better:

    tags = list(set(tags + default_tags))

jmaupetit: I think, we need to combine @k4nar and @SpotlightKid's proposals :)

```python
tags = list(set(tags or [] + default_tags))
```

watson/watson.py
```diff
@@ -425,3 +433,10 @@ def merge_report(self, frames_with_conflict):
                 merging.append(conflict_frame)
 
         return conflicting, merging
+
+    def _default_tags(self, project):
+        default_tags_raw = self.config.get('default_tags', project)
+        default_tags = []
+        if default_tags_raw is not None:
+            default_tags = shlex.split(default_tags_raw)
+        return default_tags
```
SpotlightKid: I would factor this out into a new method of the `config.ConfigParser` class.

```
    def getlist(self, section, option, default=None):
        if not self.has_option(section, option):
            return [] if default is None else default
        
        value = self.get(section, option)
        
        if '\n' in value:
            return [item for item in value.splitlines() if item]
        else:
            return shlex.split(value)
```

This also allows an alternative syntax for lists of tags, some of which have spaces in them:

```
[default_tags]
interstellar travel = solar sail
    cryogenics
    fusion
watson =
    issue 113
    python
```

watson/watson.py
jbarnoud: I would rather keep the order, which is not the case with a set. There are ordered sets in python, but not in the standard library. The simplest way to keep the order without adding a dependancy would be to write a function to deduplicate the list, which should be rather easy.

Note that the change would broader the scope of the PR. Indeed, it would deduplicate tags provided multiple time by the user in cases such as:

```
watson start project +tag + tag
```

This would be an enhancement, I think, but I prefer to warn.

SpotlightKid: Ok, another version, which preserves order (replaces line 251 - 256 and depends on #114):

```
        if not restart:
            extra_tags = self.config.getlist('default_tags', project)
            tags = extra_tags + [tag for tag in (tags or [])
                                 if tag not in extra_tags]
```

jbarnoud: I changed it for

```python
def deduplicate(sequence):
    """
    Return the input sequence without duplicate, keep the order.
    """
    return [element
            for index, element in enumerate(sequence)
            if element not in sequence[:index]]
```

and

```python
        default_tags = self.config.getlist('default_tags', project)
        if not restart:
            tags = (tags or []) + default_tags

        self.current = {'project': project, 'tags': deduplicate(tags)}
```

watson/watson.py
jbarnoud: My local version is now based on #114. I'll push when #114 get merged.

watson/watson.py
SpotlightKid: I find this overly complicated or I don't get it. Why do you look up the element only in the sublist `sequence[:index]`? If this some kind of optimization, I don't understand how it's supposed to work, and, anyway, I think creating many copies of the list defeats the purpose.

jbarnoud: I do not think that it makes sense to remove only the duplicates between the user provided tags and the default tags.  It makes much more sens to me to just deduplicate the final list of tags.

Indeed, my solution for that creates many small lists in the process. I do not think it is much of an issue as there will never be enough tags to make it slow. Though, it can get replaced by something simpler like:

```python
def deduplicate(sequence):
    result = []
    for element in sequence:
        if element not in result:
            result.append(element)
    return result
```

watson/watson.py
SpotlightKid: Ok, this took quite some time, but #114 is now merged. Please update our PR, rebase it on the current master and squash the commits. Then it can be merged.

jbarnoud: I'll take care of this latter this week.

SpotlightKid:
Ping...

jbarnoud:
Here it is. Sorry for the delay.

watson/utils.py
```diff
@@ -197,3 +197,12 @@ def safe_save(path, content, ext='.bak'):
             os.rename(path, path + ext)
 
         os.rename(tmpfp.name, path)
+
+
+def deduplicate(sequence):
+    """
+    Return the input sequence without duplicate, keep the order.
```
SpotlightKid: More accurately:

    Return a list with all items of the input sequence but duplicates removed.
    
    Keeps the order of items in the input sequence.

watson/watson.py
```diff
@@ -14,7 +14,7 @@
 
 from .config import ConfigParser
 from .frames import Frames
-from .utils import make_json_writer, safe_save
+from .utils import make_json_writer, safe_save, deduplicate
```
SpotlightKid: Minor nick-pit: keep imported names in alphabetical order.

docs/user-guide/configuration.md
```diff
@@ -142,6 +142,52 @@ Globally configure how `dates` should be formatted. All [python's `strftime` dir
 
 Globally configure how `time` should be formatted. All [python's `strftime` directives](http://strftime.org) are supported.
 
+### Default tags
+
+Tags can be automatically added for some projects. This is convenient when the
```
SpotlightKid: s/for some/to some/

Instead of "some" I would prefer "chosen", "selected", "specific" or similar.

SpotlightKid:
Thanks!

Can you address the few new comments I made, then I'll merge.

You can squash you commits or I'll do it when merging.

jmaupetit:
@jbarnoud : Hacktoberfest starts tomorrow 😉

jmaupetit:
@jbarnoud please let me know if you don't have time to finish this: I can give a hand! Once this PR has been merged, thanks to @SpotlightKid, we will release Watson 1.4.0 and prepare a RC for 2.0 (FTR, see [this comment](https://github.com/TailorDev/Watson/pull/115#issuecomment-241762562) in #115).

jbarnoud:
It get's to the top of my todo list. They are only a few items remaining 
on top of it, so it will be done soon.

jmaupetit:
@SpotlightKid what is your feeling about postponing this feature to the 2.0 release?

SpotlightKid:
The PR is almost ready. We could also just fix these minor issues I commented on above ourselves.

jbarnoud:
No need to postpone anything anymore. Sorry for the delay, my TODO list was going out of hands.

jmaupetit:
Thank you @jbarnoud ; I felt bad about fixing your stuff myself. 
@SpotlightKid I am letting you review and merge this PR, thanks!

PR #114, opened 2016-04-13
Add new method 'getlist' to 'ConfigParser'

SpotlightKid:
This allows to get a configuration value as a list of strings.

For now this mainly to support PR #113, but could be used by future new features as well.

Here's a copy of the included docstring:

```
Return value of option in given section as a list of strings.

If option is not set, return default (defaults to an empty list).

The option value is split into list tokens using one of two strategies:

* If the value contains any newlines, i.e. it was written in the
  configuration file using continuation lines, the value is split at
  newlines and empty items are discarded.
* Otherwise, the value is split according to unix shell parsing rules.
  Items are separated by whitespace, but items can be enclosed in 
  single or double quotes to preserve spaces in them.
  
Example::

    [test]
    option2 =
        one
        two three
        four
        sive six
    option1 = one  "two three" four 'five  six'
```

### DIFF
--- a/watson/config.py
+++ b/watson/config.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 """A convenience and compatibility wrapper for RawConfigParser."""
 
+import shlex
+
 try:
     from configparser import RawConfigParser
 except ImportError:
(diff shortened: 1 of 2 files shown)


k4nar:
LGTM :+1: 

It only adds what's needed know, but it can easily be extended in the future, I like it.

watson/config.py
```diff
@@ -60,6 +62,42 @@ def getboolean(self, section, option, default=False):
...
+        """
+        Return value of option in given section as a list of strings.
+
+        If option is not set, return default (defaults to an empty list).
+
+        The option value is split into list tokens using one of two strategies:
+
+        * If the value contains any newlines, i.e. it was written in the
+          configuration file using continuation lines, the value is split at
+          newlines and empty items are discarded.
```
jbarnoud: It may be worth having a test with en empty line to make sure it is discarded.

SpotlightKid: The test already covers that with `value1` in the test config, where the first option line is empty. But I will add another empty line on its own, to demonstrate that this works too.

SpotlightKid:
Is anything preventing this from getting merged? Finalizing & merging PR #113 depends on it.

watson/config.py
```diff
@@ -60,6 +62,42 @@ def getboolean(self, section, option, default=False):
...
+            option1 = one  "two three" four 'five  six'
+
+        """
+        if not self.has_option(section, option):
+            return [] if default is None else default
+
+        value = self.get(section, option)
+
+        if '\n' in value:
+            return [item for item in value.splitlines() if item]
```
k4nar: shouldn't we strip each item ? Looking at the tests it seems to work that way, but I'm not sure to understand why.

k4nar:
Sorry about the delay :( .

I made a small comment because there is something that I don't quite understand, but otherwise I'll merge it.

watson/config.py
SpotlightKid: Per the [configparser docs](https://docs.python.org/2/library/configparser.html) leading whitespace is stripped from values.

I agree it's probably best to explicitly strip items in line 97 to also strip trailing whitespace.

Just say the word and I'll update the PR.

k4nar: Yeah I think this would be a little bit clearer :) .

SpotlightKid:
I've updated the PR to strip list item values when continuation lines are used.

Issue #98, opened 2016-02-26
Edit command fails if frame id is numerical


bryanjswift:
I currently have a frame `1612702` that I'd like to edit but `watson edit 1612702` treats it as a relative frame number and I get "Error: Frame index 1612702 not in range."

SpotlightKid:
I can see [where](https://github.com/TailorDev/Watson/blob/master/watson/cli.py#L641) in the code that happens, it first tries to interpret the frame ID as an integer and assumes, if that succeeds, that it is a frame index.

We could probably fix this by using different heuristics, but actually this bug highlights issues with the way access to the internal frame database is modelled. IMO `watson.frames.Frames` tries to cram too much into the dictionary-style interface: get frame by frame ID, get frame by index and get frame data by "column" name. Since they are all handled by the method that implements the index operator (`[]`), there is currently no interface to specify whether you want to look up a frame by ID or by index.

This also makes access by frame ID rather inefficient and it gets slower as the frame database grows.

My first instinct would be to try to change the internal datatype of `Frames` into an ordered dict and provide explicit methods for access by index etc. But I'll have to take a closer look.

k4nar:
Yeah, I might have tried to be too clever here. However, if we stick with the current design, the error actually lies in the edit command: https://github.com/TailorDev/Watson/blob/master/watson/cli.py#L640-L651 . I have an idea how to fix that, I'll make a PR later today.

FIY @SpotlightKid I did a first implementation using [tablib](https://github.com/kennethreitz/tablib), but it wasn't working as I wished so I tried to do something similar but simpler. I don't have strong feelings about it, and I agree that the current design might be confusing and not very efficient in certain cases.
I guess your right, an OrderedDict might be a far better option. I'll try to give it a shot.

Issue #93, opened 2016-02-22
Add git commit-like comment on watson stop


speedRS:
This would be useful to provide historic context for a given time entry. Admittedly, this could possibly be handled via tags but I a description of "updated project estimates spreadsheet" vs. +project +estimates +update (or something similar) would make more sense.

An example of what I'm thinking is:

`$ watson start ....`
`$ watson stop ....`

Then open default editor (akin to `watson edit <frame ref>`).

Alternatively, add an additional flag `-c "some kind of comment or note"`.

SpotlightKid:
I have mixed thoughts about this proposal. A way to attach a longer note to a watson frame seems desirable. But:

- It shouldn't change the established command line interface, i.e. I should still be able to start a frame with just "watson start <project>".
- I fear that trying to do too many things with watson, i.e. in this case trying to use as a project log instead of just a time tracker, might in the long run add to much bloat and make the interface unintuitive. `git` is a particular bad example of UI design, IMHO.

jmaupetit:
Personally I like the idea to add a short report to a working session. But it shouldn't be the default behavior, @SpotlightKid made the point.

speedRS:
All good points and I agree on them all. The -c/--comment (or whatever works) would be an optional flag. The primary reason is context for my use case. There are times where all I need to track is time spent on a project; other times, we're expected to provide a brief note on what was worked on during that period for billing/justification purposes.

As some background, I'm currently oscillating between org-mode, a phyiscal notebook and Google Spreadsheets. Ideally, I don't want something as bloated as org-mode. I just like it's capabilities for time tracking and Watson seems a good replacement/fit without all the extra crap.

As I'm thinking about this, I wonder to at which point the note should/could be entered. In some respects, on clock out almost makes more sense as you will have completed the work at that point. The only issue I see then is autoclockout on starting work on a new project, unless there is say a config item to pop up an editor, get a comment, clock out and start the new clock. Again, that sounds way too git-like but it's the best example of workflow.

Anyways, happy to discuss more if you think this is something you'd be interested in.

speedRS:
Apologies. I keep click close and comment.

k4nar:
:+1: for adding messages to frames. This is actually something that Watson was doing in its first days, and then I removed it in order to focus more on tags.

Adding a `-m` option to `watson start` seems to be a good place to start. Of course messages should also show up in `watson edit` and in the log.

jmaupetit:
[...]

Issue #5, opened 2015-03-19
Add some tests for the CLI


k4nar:
Currently the lib is pretty well covered, but the CLI is not. Some functions like `start` or `edit` have some complexity in them so they should be tested.

willdurand:
IMO complexity should be extracted from the cli commands, so that it would be easier to test.

jmaupetit:
This should be a priority for the next release.

willdurand:
I can't agree more. Click lib has a test runner for that.

jmaupetit:
Just stumbled upon this: https://pypi.python.org/pypi/pytest_click/0.1 I think this might help.

PR #115, opened 2016-04-13
Refactor 'watson.frames'

SpotlightKid:
As mentioned in #98, I think the internal structure of `watson.frames.Frames` is convoluted and inefficent. The API mixes dict-style key lookup and list-style indexing and then throws in relational-style access to "column" data into the mix as well. There is currently no interface to specify whether you want to look up a frame by ID or by index. This makes access by frame ID rather inefficient and it gets slower as the frame database grows.

This PR introduces these changes:
- Made `watson.frames.Frames` a subclass of `collections.OrderedDict`.
- Changed order of items in `watson.frames.Frame` to be more consistent
  with `Frames.add()` and `Frames.new_frame()`.
  
  This changes the on-disk JSON representation of the frames, i.e. the
  format of the `<app dir>/watson/frames` file. The code handles loading
  the old format automatically and on saving the file gets written in
  the new format.
- Removed ability to look up frame by index with dict look-up syntax
  and provide `Frames.get_by_index()` method as replacement.
- Removed ability to delete frame by index (was not used aywhere).
- Removed ability to get Frames column values with dict look-up syntax
  and provide `Frames.get_column()` generator as replacement.
- Iterating over `Watson.frames` yields frame IDs instead of `Frame`
  instances. Where the latter is desired, the `values()` method must be
  used.
- Passing a `Frames` instance to the `Frames()` constructor (instead of a list) seems to have worked with the old format, but wasn't documented. This doesn't work anymore, but you can assign still a `Frames` instance to the `frames` attribute of a `watson.Watson` instance.
- Adding a frame by assigning a `Frame` instance to a key of a `Frames`,
  makes a copy of the `Frame` instance with the given key as the frame ID.
- Made `Frames.filter()` a generator.
- Adapted code using `Frame` and `Frames` to these changes.
- Adapted tests and test data files.
- The command line interface remains unchanged.


SpotlightKid:
All test pass, and I've tested the command lien interface manually under Python 2.7 and 3.5, but I think this change requires some additional testing by other contributors, to make sure there are no problems related to the format change of the `frames` file.

So if your test this, please make sure to **BACKUP YOUR WATSON DIRECTORY!**

SpotlightKid:
I optimized the access to Frames by index via the `get_by_index` method to be an order of magnitude faster than my initial implementation by maintaining a separate list of keys (frame IDs), which gets updated when adding, inserting or moving an entry.

I also added test for the `get_by_index` method to make sure the integrity of the key list is properly maintained.

SpotlightKid:
Any opinions on this?

k4nar:
I'll try to review it soon, but it looks promising :) .

SpotlightKid:
Can we make an effort to confirm this works as it should and merge it?

I want to work on #93, and I would like to base it on this, to avoid having to change it again after this gets merged.

tests/test_watson.py
```diff
@@ -716,17 +785,19 @@ def json(self):
...
+    assert frame.start.timestamp == 3
+    assert frame.stop.timestamp == 4
+    assert frame.tags == ['A']
+
+    frame = watson.frames.get_by_index(1)
+    assert frame.id == '2'
+    assert frame.project == 'bar'
+    assert frame.start.timestamp == 4
+    assert frame.stop.timestamp == 5
+    assert frame.tags == []
```
k4nar: Could we add a test to see that loading frames stored with the old format works ?

watson/frames.py
```diff
@@ -68,88 +70,73 @@ def __contains__(self, frame):
...
-        rows = [Frame(*frame) for frame in frames]
-        self._rows = rows
+        super(Frames, self).__init__()
+        self._keys = list(self.keys())
 
-        self.changed = False
+        for frame in frames or []:
+            # convert from old format with project @ idx 2 and ID @ idx 3
+            if not isinstance(frame[2], (int, float)):
+                frame = [frame[3], frame[2], frame[0], frame[1]] + frame[4:]
```
k4nar: :+1:

watson/frames.py
```diff
@@ -68,88 +70,73 @@ def __contains__(self, frame):
...
-class Frames(object):
+class Frames(OrderedDict):
     def __init__(self, frames=None):
-        if not frames:
-            frames = []
-
-        rows = [Frame(*frame) for frame in frames]
-        self._rows = rows
+        super(Frames, self).__init__()
+        self._keys = list(self.keys())
```
k4nar: For the sake of consistency, a tuple would be great there :) .

watson/frames.py
```diff
@@ -68,88 +70,73 @@ def __contains__(self, frame):
...
     def filter(self, projects=None, tags=None, span=None):
-        return (
-            frame for frame in self._rows
-            if (projects is None or frame.project in projects) and
-               (tags is None or any(tag in frame.tags for tag in tags)) and
-               (span is None or frame in span)
-        )
+        for frame in self.values():
+            if ((projects is None or frame.project in projects) and
+                    (tags is None or set(frame.tags).intersection(tags)) and
```
k4nar: I wonder if it's faster than the `any` version.

k4nar:
@jmaupetit @willdurand : Do you think that we should consider this PR as a breaking change in the API and release a 2.0?

SpotlightKid:
Just found a spot in the `merge` command, which I hadn't updated to work with the new frame format.

It seems there are no unit test for the merge command because the whole logic is lumped into the command handling function. A test might have caught this earlier.

How should we improve this situation? Include a test in this PR or open a ticket to add unit tests for the merge command?

k4nar:
We already have a ticket about adding tests for the commands (https://github.com/TailorDev/Watson/issues/5) but there haven't been any momentum on this.

I don't know if this PR is the right place to start this though.

tests/test_watson.py
SpotlightKid: Added a test in e33b6b6.

watson/frames.py
SpotlightKid: The solution using `set` is actually slower:

```
In [1]: old_tags = list("abcdefghijkl")

In [2]: new_tags = list("ghijklmnopqr")

In [3]: %timeit any(tag in old_tags for tag in new_tags)
The slowest run took 10.96 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 731 ns per loop

In [4]: %timeit set(old_tags).intersection(new_tags)
The slowest run took 10.64 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 1.15 µs per loop
```

I only used it to be able to make the line shorter and fix a stupid flake 8 complaint about indentation. Though I think the slight performance loss is acceptable in this case.

jmaupetit:
@k4nar Yes, I think we must release a major version after merging this.
@SpotlightKid I am responsible for the work (still) in progress on testing commands, there is no need to had tests for this here. That said, we will need to declare a new issue for the `merge` command and fix it before releasing a `2.0.0`.

So, if merging this PR is required for further work, LGTM! Thx @SpotlightKid 🎉

SpotlightKid:
Today I released that there is a major oversight on my part in the new Frames object from this PR. it doesn't support frame lookup by ID prefix. This causes commands like `watson edit <frame ID>` not to work anymore. :( (Another strong indication that watson needs test for the cli commands.)

I'll have to research if there's a solution to add support for this without hurting performance too much.

SpotlightKid:
Added frame lookup by ID prefix with the simplest approach that works, i.e. looping over the keys. To optimize for the most likely use, i.e. looking up recently added frames, the loop iterates over the IDs in reverse order (of addition). The behaviour thus differs from original implementation in that for keys with a common prefix the last added is found first and it value returned.

I've used watson with this new storage extensively now for a few weeks and except for this ID prefix issue, had no problems.

k4nar:
It looks great. I would like to take a last look at it before merging it, but I'm not when I'll have time for that. Maybe next week :) .

jmaupetit:
@k4nar if you have time, I'll let you review this PR since it's related to the core part you designed first. Please, ping me if you don't.

k4nar:
LGTM :+1:

SpotlightKid:
I would like to merge #113 (when its ready), release 1.4.0, then merge this and #119 and prepare a release candidate for 2.0.

jmaupetit:
Sounds good to me. Thx!

SpotlightKid:
Closing this since I have no interest anymore in getting this merged.

If anyone is still interested in this, feel free to open a new PR based on my commits.

PR #117, opened 2016-05-23
Add regex in project name for frame filter

yloiseau:
This PR adds support for regular expressions to the `--project` option.


yloiseau:
Two questions:
- should regex be used or glob ([`fnmatch`](https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatch))?
    - regex are more powerful
    - glob are simpler to use

- should the feature be activated by a configuration option?

k4nar:
That's a great idea, but I don't really like having it as a default. It could create some confusion or weird errors.

Maybe we could have another argument for that, like `-r` ? It could even be a flag activating the same thing for tags.

yloiseau:
I was indeed planning to add it to tags, that would allow negative selectors.
I think a `--regex` flag turning it on for both project and tags filters would be the more natural, with a config option to make it default.

yloiseau:
And while we are parametrizing, another flag to use glob instead of regex :smile:

k4nar:
Your question "should we use glob or regex" is still valid though :) . Maybe globs are enough.
I can't really think of a use case where I would need something beyond `-p foo-*` or `-p *-bar`.

yloiseau:
A use case for real regex could be to match frames not matching a criteria, for instance:
```
watson log -p 'Project\.(?!Subproject)'
```

docs/user-guide/commands.md
```diff
@@ -340,6 +341,7 @@ Flag | Help
 `-t, --to DATE` | The date at which the report should stop (inclusive). Defaults to tomorrow.
 `-p, --project TEXT` | Reports activity only for the given project. You can add other projects by using this option several times.
 `-T, --tag TEXT` | Reports activity only for frames containing the given tag. You can add several tags by using this option multiple times
+`-m, --matcher TEXT` | The matching method to use when filtering frames. Can be `regex`, `glob`, of `fixed` (the default).
```
jmaupetit: `s/of/or`

watson/cli.py
```diff
@@ -414,8 +419,11 @@ def report(watson, from_, to, projects, tags):
               help="Logs activity only for frames containing the given "
               "tag. You can add several tags by using this option multiple "
               "times")
+@click.option('-m', '--matcher', 'matcher',
+              help="The matching method to use when filtering frames. "
+              "Can be `regex`, `glob`, of `fixed` (the default).")
```
jmaupetit: `s/of/or`

docs/user-guide/configuration.md
```diff
@@ -146,6 +154,7 @@ stop_on_start = true
 stop_on_restart = false
 date_format = '%Y.%m.%d'
 time_format = '%H:%M:%S%z'
+matcher = regex
```
jmaupetit: `regex` shouldn't be `'regex'`?

yloiseau: Actually since it's just a string, it's OK without the quotes

.gitignore
```diff
@@ -26,3 +26,4 @@ node_modules/
 .tox
 dist/
 site/
+.env
```
SpotlightKid: This change shouldn't really be in this PR, should it?

docs/user-guide/commands.md
SpotlightKid: I find the term "fixed" a bit unintuitive. How about "exact"?

docs/user-guide/configuration.md
SpotlightKid: Actually, I wonder why the `date_format` and `time_format` settings have quotes in this example. AFAICS they shouldn't.

docs/user-guide/commands.md
yloiseau: It's a reference to `grep` (`fgrep`/`-F|--fixed-regexp`)

SpotlightKid: Ah, ok, I see.

docs/user-guide/configuration.md
yloiseau: Should I make the change in this PR ?

.gitignore
yloiseau: You mean that the ignore should not be present, or that a separate PR should be made for this change only?

SpotlightKid: Separate PR or at least not in this one, IMHO.

watson/cli.py
```diff
@@ -414,8 +419,11 @@ def report(watson, from_, to, projects, tags):
               help="Logs activity only for frames containing the given "
               "tag. You can add several tags by using this option multiple "
               "times")
+@click.option('-m', '--matcher', 'matcher',
```
SpotlightKid: The `-m` option is now already used as the short form of `--month` and conflicts with the usage here.

SpotlightKid:
I'm wondering whether it is really necessary to expose the matcher type as a command line option. IMHO having it as config option is enough. I don't think this is something you would want to change often on-the-fly.

The `report` and `log` commands already have too many options now as is.

watson/utils.py
```diff
@@ -111,3 +114,20 @@ def get_frame_from_argument(watson, arg):
             style('error', "No frame found with id"),
             style('short_id', arg))
         )
+
+
+def name_matcher(kind):
+    """
+    Create a function to match a name with an expression.
+    Used in frame filters to select projects and tags.
```
SpotlightKid: Please insert an empty line here.

watson/cli.py
```diff
@@ -15,7 +15,7 @@
 from . import watson
 from .frames import Frame
 from .utils import (format_timedelta, get_frame_from_argument, options,
-                    sorted_groupby, style)
+                    sorted_groupby, style, name_matcher)
```
SpotlightKid: Please sort imported names alphabetically.

SpotlightKid:
Can you please:

* Rebase on the current master (and fix conflicts).
* Remove the change to `.gitignore` (already in master in another form)
* Work in other annotations.
* Decide what to do with the command line option and, if you keep it, rename the short form.
* Squash commits.

SpotlightKid:
Just noticed an issue with the `report` command when filtering frames by tags with the regex or glob matcher:

{{{
$ watson report -y -p watson
Fri 01 January 2016 -> Sun 31 July 2016

watson - 44m 27s
	[frames          07m 12s]
	[issue-93        16m 15s]
	[logo            21m 00s]
	[notify          21m 00s]
	[refactoring     07m 12s]

$ watson report -y -M regex -p watson -T 'issue-.*'
Fri 01 January 2016 -> Sun 31 July 2016

watson - 16m 15s
}}}

The selected tags are not listed under the project. The issue is in how the `tags_to_print` set is calculated:

    tags_to_print = sorted(
            set(tag for frame in frames for tag in frame.tags
                if tag in tags or not tags)
    )

This uses `tags` but should instead should use the result of filtering the tags by the given regex or glob pattern.

I would also strongly suggest not to use a nested generator expression. They are generally very hard to parse mentally for others reading of the code.

yloiseau:
@SpotlightKid I disagree on the need to change the matcher kind on-the-fly. As an analogy, I often change `grep` expression type (`-G` or `grep`, `-E` or `egrep`,  and `-F`  or `fgrep`) depending on what I’m looking for.

yloiseau:
BTW, I did not introduced the nested generator expression :smile: (while I don't find them harder to read than nested loops)

SpotlightKid:
Yes, it was already there, but it needs to changed anyway and probably `tags` needs to be replaced by the result o fa function call further up. I would find it a lot clearer, to unravel the nesting here.

The problem I have with nested generator expressions or list comprehensions is that it is not clear (without knowing the parsing rules) from the syntax, which is the outer and which is the inner loop, and that the item expression at the beginning is separated from the for expression defining it by the outer for expression. In my opinion that qualifies as hard to read.

yloiseau:
As a side note, my changes to `.gitignore` are not in master in another form, since I don't ignore the virtualenv configuration, but the autoenv one.

SpotlightKid:
I see. I pushed a cleanup of the `.gitignore`  file in #131, which includes `.env`.

PR #120, opened 2016-06-11
Handle errors when saving files gracefully

### LINKED ISSUE #118: Make backup before writing to prevent possible loss of frames file
During development I accidentally introduced an error in `frames.Frames.dump()`, which wasn't caught in `watson.Watson.save()` and resulted in my frames file being wiped, i.e. ending up as a zero-length file (I had a backup of course).

I think we should do the following:

* Write the new frames file to a different file first and move it to the destination only if the write was successful.
* Keep the previous file as a backup.

We could move the existing file out of the way first, but then we would need to revert this, if the write fails.

SpotlightKid:
Fixes #118.

* Prevent existing files from being wiped when writing fails
* Backup previous file if it exists when writing succeeds

To fix test failures with python 2.7 I had to change the imports at the top of the tests so that the test under Python 2.7 always use `StringIO.StringIO` instead of `io.StringIO`. The latter doesn't cooperate with `json.dump` under Python 2.7.

### DIFF
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -9,9 +9,9 @@ except ImportError:
     import mock
 
 try:
-    from io import StringIO
-except ImportError:
     from StringIO import StringIO
+except ImportError:
+    from io import StringIO
 
 import py
 import pytest
@@ -23,7 +23,8 @@ from dateutil.tz.tz import tzutc
 from click import get_app_dir
 from watson import Watson, WatsonError
 from watson.watson import ConfigurationError, ConfigParser
-from watson.utils import get_start_time_for_period
+from watson.utils import get_start_time_for_period, make_json_writer, safe_save
+
 
 TEST_FIXTURE_DIR = py.path.local(
     os.path.dirname(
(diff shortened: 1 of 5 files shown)


tests/test_watson.py
```diff
@@ -8,9 +8,9 @@
     import mock
 
 try:
-    from io import StringIO
-except ImportError:
     from StringIO import StringIO
+except ImportError:
+    from io import StringIO
```
k4nar: :+1:, `io.StringIO` doesn't have the right behavior in Python 2.7. This is very confusing.

watson/utils.py
```diff
@@ -111,3 +114,44 @@ def get_frame_from_argument(watson, arg):
             style('error', "No frame found with id"),
             style('short_id', arg))
         )
+
+
+def make_json_writer(func, *args, **kwargs):
```
k4nar: I don't get it, why expecting a function and not an object?

SpotlightKid: I wanted to handle the two cases of `config.write(fp)` and `json.dump(data, fp)` with a common interface, so `safe_save` expects a function, which takes a file-object and knows how to write some data to it. This can be `config.write`, for example, or the function returned by `make_json_writer`.

I wanted to avoid having to store the data, which the function writes, in memory, so instead of passing the data to `make_json_writer`, you pass a function, which generates the data to write, for example `frames.dump`. The effect is that the function, which generates the data, is only called at the moment the writer function passed to `safe_save` is called. If the data generation function is a e.g. a generator and the writer function can take advantage of it, this makes writing more efficient than passing the whole lump of data as a variable.

PR #125, opened 2016-06-15
Improve make facility for developer's environment creation

alexkey:
- Non-build targets must be `.PHONY` – corrected.
- `PYTHON` and `PIP` variables from an environment; usage: `PYTHON=python3 make env`.
- Additional `check` and `clean` targets added for a convenience.
- New make targets `$(VENV_DIR)` and `env` added. These targets are suitable for creation of the isolated developer's environment within application directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ node_modules/
 .tox
 dist/
 site/
+.env/
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,7 @@
 # Watson
 
-ifndef PYTHON
-	PYTHON = python
-endif
-
-ifndef PIP
-	PIP = pip
-endif
+PYTHON ?= python
+PIP ?= pip
 
 VENV = virtualenv
 VENV_ARGS = -p $(PYTHON)
--- a/Makefile
+++ b/Makefile
@@ -1,21 +1,54 @@
 # Watson
 
+ifndef PYTHON
+	PYTHON = python
+endif
+
+ifndef PIP
+	PIP = pip
+endif
+
+VENV = virtualenv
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -1,21 +1,54 @@
 # Watson
 
+ifndef PYTHON
+	PYTHON = python
```
k4nar: Isn't it the same thing than `PYTHON ?= python`

alexkey: Yes, this is a short form – corrected.

Makefile
```diff
@@ -1,21 +1,49 @@
 # Watson
 
+PYTHON ?= python
+PIP ?= pip
+
+VENV = virtualenv
+VENV_ARGS = -p $(PYTHON)
+VENV_DIR = $(CURDIR)/.env
```
yloiseau: I think another name should be chosen for the venv dir (`.venv`?) since `.env` conflicts with the [`autoenv`](https://github.com/kennethreitz/autoenv) file

alexkey: Done.

.gitignore
```diff
@@ -26,3 +26,4 @@ node_modules/
 .tox
 dist/
 site/
+.env/
```
SpotlightKid: Should be `.venv/`.

SpotlightKid:
Can you make the change to `.gitignore` I just commented on and then squash your four commits into one (e.g. with `rebase -i HEAD~4`)? I'll then do the merge right after.

SpotlightKid:
I went ahead and created a new PR 129 based on this, fixed the `.gitignore` file, squashed the commits and merged it. So I'm closing this one. Thanks for the Makefile improvement!

PR #133, opened 2016-08-13
Suggest to use tox by default, describe how to run tests

SpotlightKid:
Also sets the tox config option [skip_missing_interpreters](https://testrun.org/tox/latest/config.html#confval-skip_missing_interpreters=BOOL) to `True`, so tox can be run even you don't have all the Python versions defined in `tox.ini` installed.

Lastly, adds a note in the PR guidelines about marking WIP PRs with "WIP:".

Replaces #104.


jmaupetit:
LGTM

k4nar:
LGTM :+1:

Issue #135, opened 2016-10-01
utils.safe_save fails when tmpdir is on another filesystem


sazarkin:
Could not start project without this adjustment.

```
$ ➜ watson start test
Starting project test  at 21:34
Error: Impossible to write None: [Errno 18] Invalid cross-device link
```

OS: ubuntu 16.04 x64

SpotlightKid:
This is probably because the temporary directory is on another partition than the watson dir.

PR #136, opened 2016-10-01
Replace os.rename with higher-level shutil.move

sazarkin:
Issue #135

### DIFF
--- a/watson/utils.py
+++ b/watson/utils.py
@@ -2,6 +2,7 @@ import datetime
 import itertools
 import json
 import os
+import shutil
 import tempfile
 
 import click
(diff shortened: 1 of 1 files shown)


jmaupetit:
Oops.The circle-ci integration wasn't expected for this repository. Sorry for that.

k4nar:
Nice catch !

SpotlightKid:
Here's a test, which will simulate that the temp file being used by `safe_save()` is on another filesystem. The test will fail without the patch and pass with it.

I'm not adding it to the PR yet, since I want PR #139 to be merged first, otherwise there will be a merge conflict, if I add the test into `test_watson.py` now.

```python
def test_safe_save_tmpfile_on_other_filesystem(config_dir, mock):
    save_file = os.path.join(config_dir, 'test')
    backup_file = os.path.join(config_dir, 'test' + '.bak')

    assert not os.path.exists(save_file)
    safe_save(save_file, lambda f: f.write("Success"))
    assert os.path.exists(save_file)
    assert not os.path.exists(backup_file)

    with open(save_file) as fp:
        assert fp.read() == "Success"

    # simulate tmpfile being on another file-system
    # OSError is caught and handled by shutil.move() used by save_safe()
    mock.patch('os.rename', side_effect=OSError)
    safe_save(save_file, "Again")
    assert os.path.exists(backup_file)

    with open(save_file) as fp:
        assert fp.read() == "Again"
```

PR #139, opened 2016-10-30
Refactor tests for different modules, fixtures and test utils into separate modules

SpotlightKid:
The existing tests where all in one file and this was getting way too long.

So I put tests for the the `watson.config` and `watson.utils` modules into their own `test_*.py` files and created a `conftest.py` file for the test fixtures and put some test utility functions in a `__init__.py`module for the `test` package dir.

I also changed the tests to use the `mock` fixture provided py `pytest-mock`. This makes the tests easier to read and safeguards against patching side-effects without having to use `with` everywhere. Lastly, I redefined the `_dt` helper function used in `test_utils.py` using `functools.partial` to save some repetition.

Signed-off-by: Christopher Arndt chris@chrisarndt.de

PR #140, opened 2016-10-30
Make formatting styles configurable

SpotlightKid:
Styles can be set via sections in the config file named `style:<element>`, where `<element>` is the name of an element a style can be applied to. Settings in that section are used as parameters to the `click.style` function.

The current list of possible elements is:

- project
- tag
- time
- error
- date
- id

Example config:

```
[style:project]
fg = white
bg = black

[style:tag]
fg = yellow
bg = white
```

Notes:
- This PR depends on PR #139 
- Documentation is not updated yet, it will be added when the reviewers agree this is a good config structure and implementation.
- I had originally contemplated to use a config structure as shown below or similar, but it would have made the config parsing much more complicated.

```
[styles]
project = fg:white bg:black
tag = fg:red
...
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 arrow
-click
+click>=5.0
 requests
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -144,6 +144,21 @@ value5 = one
         "Modifying default return value should not have side effect.")
 
 
+def test_config_getitems(mock, watson):
+    content = """
+[test1]
+foo=bar
+spamm=eggs
+
+[test2]
+"""
+    mock.patch.object(ConfigParser, 'read', mock_read(content))
+    assert watson.config.getitems('test1') == dict(foo='bar', spamm='eggs')
+    assert watson.config.getitems('test2') == {}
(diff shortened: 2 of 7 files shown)


k4nar:
Neat :+1:!

Although, may I suggest the following format instead?
```
[style]
project_fg = white
project_bg = black
tag_fg = red
```

SpotlightKid:
I'm not sure, if I like that one better. Makes the options rather verbose and if you don't keep options for one element together, it might become harder to see all options for one element at a glance.

k4nar:
Other proposal, very close to yours:
```
[style:tag]
fg = yellow
bg = white
```

I don't think `:` is part the ini "spec", but I've seen it used before. For example this is what Circus uses.

SpotlightKid:
That would be a single-char change in the code (plus adapting tests).

k4nar:
I guess so yes. That's just a matter of taste, but I find it clearer and more beautiful :) .

SpotlightKid:
Closing this since I'm cleaning my unmerged/unfinished PRs and this hasn't been acted upon in years.

PR #151, opened 2017-05-04
[WIP] Print log to terminal if option selected or if not much text

mwiens91:
The way Watson currently behaves it's quite difficult to edit frames by hash: to get the hash of a frame AND enough context to figure out what that frame is about you have to type in 'watson log'; but this brings up a pager and when you close this pager all of the log information vanishes, making it hard to remember the hash number to do a 'watson edit hash_here'.

It would be much easier if there was an option to print the log to the terminal. So I added this option. I also made it so that if the log text can fit in the terminal window, then it should print to the terminal (not the pager); this is how 'git log' works.

To determine the terminal size I used the 'stty size' command over a popen, as recommended here: https://stackoverflow.com/questions/566746/how-to-get-linux-console-window-width-in-python. Apparently this command is widely supported.

### DIFF
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -607,7 +607,14 @@ def log(watson, current, from_, to, projects, tags, year, month, week, day):
             for frame in frames
         ))
 
-    click.echo_via_pager('\n'.join(lines))
+    num_term_rows, _ = [int(i)
+                        for i in os.popen('stty size', 'r').read().split()]
+    num_rows = len(watson.frames.dump()) + (len(lines)+1) / 3 * 2 - 1
+
+    if num_term_rows > num_rows:
+        click.echo('\n'.join(lines))
+    else:
+        click.echo_via_pager('\n'.join(lines))
 
 
 @cli.command()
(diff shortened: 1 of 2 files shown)


mwiens91:
I just realized that len(lines) didn't do what I thought it did. I need to fix this. I will rebase and comment when this is fixed.

watson/cli.py
```diff
@@ -607,7 +610,14 @@ def log(watson, current, from_, to, projects, tags, year, month, week, day):
             for frame in frames
         ))
 
-    click.echo_via_pager('\n'.join(lines))
+    num_term_rows, _ = [int(i)
+                        for i in os.popen('stty size', 'r').read().split()]
+    num_rows = len(watson.frames.dump()) + (len(lines)+1) / 3 * 2 - 1
```
mwiens91: This is a very hacky and un-elegant way of determining the number of rows Watson should print to the terminal. I'm going to have to find a better way to do this. Open to suggestions if anyone sees an easy way to determine the number of rows.

mwiens91:
Likely not going to have time to work on this for the next while.

jmaupetit:
Ok then. Do not hesitate to re-open this PR if you have spare time @mwiens91!

PR #152, opened 2017-06-07
Make the sync command works with the new Crick API

jmaupetit:
## Purpose

As introduced in our [recent blog post](https://tailordev.fr/blog/2017/06/07/le-lab-5-crick-a-backend-for-watson-time-tracker/), we built a new back-end for Watson, called Crick. For now, there is no public release of the source code, but it will be open sourced soon. In the meantime, we published the specifications of its API at: http://docs.crickapi.apiary.io.

Given this new back-end API, we had to make the `sync` command work again. There won't be any side effect as AFAIK no one used it in the real world (even us), since it was not completely wired anyway.

## Proposal

This PR fixes Watson's `pull` and `push` methods to properly consume the new API.

### DIFF
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -518,9 +518,9 @@ def test_push(mock, watson):
     watson.frames.add('bar', 3, 4)
 
     mock.patch.object(watson, '_get_remote_projects', return_value=[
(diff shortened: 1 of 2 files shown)


willdurand:
Looks good to me 👍

PR #155, opened 2017-07-31
Bump release to 1.5.0

jmaupetit:
* Added: the `report` command now supports JSON output (#102)
* Updated: the `sync` command is now compatible with the new crick.io backend
  API (#152)
* Updated: Python 3.6 is now officially supported (#150)
* Fixed: catch error when user wants to edit config but file does not exist yet
  (#154)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ CHANGELOG
 This document records all notable changes to Watson. This project adheres to
 [Semantic Versioning](http://semver.org/).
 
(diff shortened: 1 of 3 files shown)


willdurand:
Awesome! We should update the `homebrew` formula I guess?

jmaupetit:
Yup. My PR is waiting: https://github.com/Homebrew/homebrew-core/pull/16240

jmaupetit:
I'll had a detailed documentation on how to release Watson.

PR #159, opened 2017-08-01
Fix incomplete renaming of  variable (fixes #158)

SpotlightKid:
Signed-off-by: Christopher Arndt <chris@chrisarndt.de>

Issue #158, opened 2017-08-01
Watson config throwing errors


rejinka:
Watson version: 1.5.0

While using the example from the docs (https://tailordev.github.io/Watson/user-guide/configuration/#editing), i get following error:

```
$ watson config backend.token 7e329263e329
Traceback (most recent call last):
  File "/usr/local/bin/watson", line 11, in <module>
    load_entry_point('td-watson==1.5.0', 'console_scripts', 'watson')()
  File "/usr/local/Cellar/watson/1.5.0/libexec/lib/python2.7/site-packages/pkg_resources/__init__.py", line 564, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/local/Cellar/watson/1.5.0/libexec/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2652, in load_entry_point
    return ep.load()
  File "/usr/local/Cellar/watson/1.5.0/libexec/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2306, in load
    return self.resolve()
  File "/usr/local/Cellar/watson/1.5.0/libexec/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2312, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/local/Cellar/watson/1.5.0/libexec/lib/python2.7/site-packages/watson/__main__.py", line 6, in <module>
    cli.cli()
  File "/usr/local/Cellar/watson/1.5.0/libexec/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/Cellar/watson/1.5.0/libexec/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/Cellar/watson/1.5.0/libexec/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/Cellar/watson/1.5.0/libexec/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/Cellar/watson/1.5.0/libexec/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
[...]

PR #161, opened 2017-08-02
Fix watson config command

jmaupetit:
New configuration was never saved on disk.

Follow up of #159 & #158


jmaupetit:
We'll need to release a `1.5.2`. As @SpotlightKid already said, we strongly need to test commands. I need to work on this.

PR #166, opened 2017-08-04
added optional flags to filter status call

McKean:
First of all thanks for making watson, I really like it but noticed that I don't use it that often...
In order to fix this I added the status of watson to my statusbar (i3). In order to not make some messy string manipulations I decided to add some optional flags for status... so in the statusbar config you can do something like this:
`dat="$(watson status -p) $(watson status -e) |"`
that way I get a nice display of what I'm currently working on and a message if I'm not tracking anything.

I'm not a python dev, so feel free to clean things up.

### DIFF
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -249,8 +249,14 @@ def cancel(watson):
 
 
 @cli.command()
+@click.option('-p', '--project', is_flag=True,
+              help="only output project")
+@click.option('-t', '--tags', is_flag=True,
+              help="only show tags")
+@click.option('-e', '--elapsed', is_flag=True,
+              help="only show time elapsed")
 @click.pass_obj
-def status(watson):
+def status(watson, project, tags, elapsed):
     """
     Display when the current project was started and the time spent since.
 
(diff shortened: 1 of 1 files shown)


willdurand:
Looks nice. I have personally patched Pure (ZSH theme): https://github.com/willdurand/dotfiles/commit/a6b2e84cb7c1199b2e3a355d160a627ff41b6732 and created a VIM plugin for airline: https://github.com/willdurand/dotfiles/blob/master/vim/bundle/watson/plugin/watson.vim.

McKean:
very nice! So instead of listing what the actual project is you just make sure that you're tracking. Also a very nice aproach, stil hope this pull request gets through so I can toss it all over the place :)

k4nar:
Good idea 👍 .

McKean:
sorry to bump this, but are you still considering pull requests? I'm holding off updating as I did this changes to my local installation and integrated it nicely into my i3-bar. I'd prefer to keep the updates coming so if PR's ain't really being merged then I'd find another way.  
Thanks

PR #167, opened 2017-08-22
Add --json option to log command

rianmcguire:
Similar to `report --json`, this makes it easier to integrate Watson with other tools.

Example output:
```json
[
    {
        "id": "9acb66330c8a41fe96a86247d2457f84",
        "project": "myproject",
        "start": "2017-08-22T07:51:00+08:00",
        "stop": "2017-08-22T08:05:17+08:00",
        "tags": [
            "dev"
        ]
    },
    {
        "id": "a06a547e546848cd90253994836b45f3",
        "project": "myproject",
        "start": "2017-08-22T11:00:05+08:00",
        "stop": "2017-08-22T11:12:09+08:00",
        "tags": [
            "invoicing"
        ]
    }
]
```

I had to tweak the existing code a little to stay within the flake8 line limit.

### DIFF
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -502,8 +502,11 @@ def report(watson, current, from_, to, projects,
               help="Logs activity only for frames containing the given "
               "tag. You can add several tags by using this option multiple "
               "times")
+@click.option('-j', '--json', 'format_json', is_flag=True,
+              help="Format the log in JSON instead of plain text")
 @click.pass_obj
-def log(watson, current, from_, to, projects, tags, year, month, week, day):
+def log(watson, current, from_, to, projects, tags, year, month, week, day,
+        format_json):
     """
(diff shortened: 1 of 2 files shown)


watson/cli.py
```diff
@@ -564,50 +567,66 @@ def log(watson, current, from_, to, projects, tags, year, month, week, day):
...
+                'id': frame.id,
+                'start': frame.start.isoformat(),
+                'stop': frame.stop.isoformat(),
+                'project': frame.project,
+                'tags': frame.tags,
+            }
+            for frame in filtered_frames
+        ]
+        click.echo(json.dumps(log, indent=4, sort_keys=True))
+    else:
```
k4nar: I think it would be cleaner to return in the if, so that we can avoid the else and the indentation.

PR #175, opened 2017-12-09
Fix setup.py

MinchinWeb:
Tried to install the project (`pip install -e .`) and it broke with the following error:

> 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers; Unordered types are not allowed

This pull request is a one-line change to return the requirements as a tuple rather than a set, and thus allow the project to install.

I'm on Python 3.6.3 on Windows 10 x64.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ def parse_requirements(requirements, ignore=('setuptools',)):
             pkg = line.strip()
             if pkg not in ignore:
                 packages.add(pkg)
-        return packages
+        return tuple(packages)
 
 
 setup(
--- a/watson/utils.py
+++ b/watson/utils.py
@@ -183,7 +183,7 @@ def safe_save(path, content, ext='.bak'):
                 tmpfp.write(content)
             else:
                 content(tmpfp)
-    except:
+    except Exception:
         try:
             os.unlink(tmpfp.name)
         except (IOError, OSError):


SpotlightKid:
Uh, when did pip change in this manner?

jmaupetit:
I encountered the same behavior recently.

jmaupetit:
Looks like we also have to fix (or ignore) this new `E722` flake8 rule.

MinchinWeb:
I could fix the flake8 error simply enough.

The remaining Travis-CI failure is Python 3.3, because [pytest dropped support](https://docs.pytest.org/en/latest/changelog.html#pytest-3-3-0-2017-11-23) with version 3.3.0, released about 2 weeks ago. Should support for Python 3.3 be dropped here too? Alternately, *pytest* could be pinned to v3.2.5, but I don't know enough about tox to pull that off myself.

SpotlightKid:
> Alternately, pytest could be pinned to v3.2.5, but I don't know enough about tox to pull that off myself.

Changing line 7 in `tox.ini` to:

```
    pytest<=3.2.5
```

should do the trick.

Unfortunately there's also a problem with `arrow` and Python 2.7. Arrpow release 0.11.0 and 0.12.0 have broken dependencies for Python 2.7 (`backports.functools_lru_cache` is not declared). So we have to also put 

    arrow<0.11.0

in `requirements.txt` for all tests to pass when running tox (without any existing virtualenv).

MinchinWeb:
@SpotlightKid see #180 for an alternate way of dealing with the requirement issue.

winduptoy:
This is preventing me from installing the AUR version, it just fails with `'install_requires' must be a string or list of strings containing valid project/version requirement specifiers; Unordered types are not allowed`.

Mind publishing a new release if this fixed it?

PR #176, opened 2017-12-09
Coloured log output

MinchinWeb:
At least on Windows `click.echo_through_pager()` seems to eat the ANSI colour codes. This PR "fixes" that by allowing the pager to be bypassed and use the "regular" `click.echo()` (which displays colour codes fine). It does this by adding a `--pager/--no-pager` option to the *log* sub-command, and a corresponding `options.log_pager` configuration setting.

image

It also makes a couple other small changes:

- requires/includes PR #175 (I can't install *watson* otherwise...)
- changes the formatting slightly of the log's daily heading to better match other output
- removes an extra space that is displayed several places is a frame has no corresponding tags.

An issue has been submitted upstream to click (see https://github.com/pallets/click/issues/901) but this seems like a practical work-around.

Also, the main image on the [first page of the documentation](https://tailordev.github.io/Watson/) should be updated. It purportedly demonstrates the *report* sub-command, but the output looks like it's actually the *log* sub-command.

Thirdly, is there a comprehensive list of configuration file settings somewhere in the documentation?

Lastly, thanks for making *watson*; I'm super excited to have discovered it!

### DIFF
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -128,7 +128,7 @@ and `--year`, `--month` and `--week` to the current year, month or week
 respectively.
 
 If you are outputting to the terminal, you can selectively enable a pager
-through the `--pager`` option.
+through the `--pager` option.
 
 You can limit the log to a project or a tag using the `--project` and
 `--tag` options. They can be specified several times each to add multiple
@@ -177,7 +177,7 @@ Flag | Help
 `-p, --project TEXT` | Logs activity only for the given project. You can add other projects by using this option several times.
 `-T, --tag TEXT` | Logs activity only for frames containing the given tag. You can add several tags by using this option multiple times
 `-j, --json` | Format the log in JSON instead of plain text
-`-v, --pager / -V, --no-pager` | (Don't) view output through a pager.
+`-g, --pager / -G, --no-pager` | (Don't) view output through a pager.
 `--help` | Show this message and exit.
 
 ## `merge`
@@ -334,7 +334,7 @@ You can limit the report to a project or a tag using the `--project` and
 projects or tags to the report.
 
 If you are outputting to the terminal, you can selectively enable a pager
-through the `--pager`` option.
+through the `--pager` option.
 
 You can change the output format for the report from *plain text* to *JSON*
 by using the `--json` option.
@@ -420,7 +420,7 @@ Flag | Help
 `-p, --project TEXT` | Reports activity only for the given project. You can add other projects by using this option several times.
 `-T, --tag TEXT` | Reports activity only for frames containing the given tag. You can add several tags by using this option multiple times
 `-j, --json` | Format the report in JSON instead of plain text
-`-v, --pager / -V, --no-pager` | (Don't) view output through a pager.
+`-g, --pager / -G, --no-pager` | (Don't) view output through a pager.
 `--help` | Show this message and exit.
 
 ## `restart`
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -127,6 +127,9 @@ You can also use special shortcut options for easier timespan control:
 and `--year`, `--month` and `--week` to the current year, month or week
 respectively.
 
+If you are outputting to the terminal, you can selectively enable a pager
+through the `--pager`` option.
+
 You can limit the log to a project or a tag using the `--project` and
 `--tag` options. They can be specified several times each to add multiple
 projects or tags to the log.
@@ -174,6 +177,7 @@ Flag | Help
 `-p, --project TEXT` | Logs activity only for the given project. You can add other projects by using this option several times.
 `-T, --tag TEXT` | Logs activity only for frames containing the given tag. You can add several tags by using this option multiple times
 `-j, --json` | Format the log in JSON instead of plain text
+`-v, --pager / -V, --no-pager` | (Don't) view output through a pager.
 `--help` | Show this message and exit.
 
 ## `merge`
@@ -329,6 +333,9 @@ You can limit the report to a project or a tag using the `--project` and
 `--tag` options. They can be specified several times each to add multiple
 projects or tags to the report.
 
+If you are outputting to the terminal, you can selectively enable a pager
+through the `--pager`` option.
+
 You can change the output format for the report from *plain text* to *JSON*
 by using the `--json` option.
 
@@ -413,6 +420,7 @@ Flag | Help
 `-p, --project TEXT` | Reports activity only for the given project. You can add other projects by using this option several times.
 `-T, --tag TEXT` | Reports activity only for frames containing the given tag. You can add several tags by using this option multiple times
 `-j, --json` | Format the report in JSON instead of plain text
+`-v, --pager / -V, --no-pager` | (Don't) view output through a pager.
 `--help` | Show this message and exit.
 
 ## `restart`
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -352,7 +352,7 @@ _SHORTCUT_OPTIONS = ['year', 'month', 'week', 'day']
               "times")
 @click.option('-j', '--json', 'format_json', is_flag=True,
               help="Format the report in JSON instead of plain text")
-@click.option('-v/-V', '--pager/--no-pager', 'pager', default=None,
+@click.option('-g/-G', '--pager/--no-pager', 'pager', default=None,
               help="(Don't) view output through a pager.")
 @click.pass_obj
 def report(watson, current, from_, to, projects,
@@ -377,7 +377,7 @@ def report(watson, current, from_, to, projects,
     projects or tags to the report.
 
     If you are outputting to the terminal, you can selectively enable a pager
-    through the `--pager`` option.
+    through the `--pager` option.
 
     You can change the output format for the report from *plain text* to *JSON*
     by using the `--json` option.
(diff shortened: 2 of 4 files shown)


SpotlightKid:
> Thirdly, is there a comprehensive list of configuration file settings somewhere in the documentation?

https://tailordev.github.io/Watson/user-guide/configuration/

SpotlightKid:
> changes the formatting slightly of the log's daily heading to better match other output
> removes an extra space that is displayed several places is a frame has no corresponding tags.

Can you please put these into a separate PR (they can go together)?

watson/cli.py
```diff
@@ -833,8 +835,8 @@ def remove(watson, id, force):
 
     if not force:
         click.confirm(
-            "You are about to remove frame "
-            "{project}{tags_spacer}{tags} from {start} to {stop}, continue?".format(
+            'You are about to remove frame {project}{tags_spacer}{tags} from '
```
SpotlightKid: Thanks for your effort in unifying single/double quotes, but for messages shown to the user, which are (in theory) translatable, we prefer double quotes. This usage might not be consistent in the existing code, but for new code please adhere to this convention.

SpotlightKid:
> Also, the main image on the first page of the documentation should be updated. 

See #177.

watson/cli.py
MinchinWeb: Fixed in https://github.com/TailorDev/Watson/pull/179/commits/903e75fafdbd51b8ea2f04f087fcd2d0e1f13534 (part of #179)

MinchinWeb:
Ok, the other changes have been split off into separate PR's, and everything here just has to do with making the pager on the log output option. Documentation has also been added.

SpotlightKid:
I see that the output of the `report` command is currently not sent through a pager. I think we should do the same there as for the `log` command and use a config option name that fits for both, e.g. `use_pager`.

SpotlightKid:
`docs/user-guide/commands.md` also needs to be updated by running `scripts/gen-cli-docs.py`. You should also update the docstrings of the `log` and `report` functions and mention the new option.

watson/cli.py
```diff
@@ -529,9 +529,11 @@ def report(watson, current, from_, to, projects,
               "times")
 @click.option('-j', '--json', 'format_json', is_flag=True,
               help="Format the log in JSON instead of plain text")
+@click.option('--pager/--no-pager', 'pager', default=None,
```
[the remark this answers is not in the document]
MinchinWeb: -v/-V (for 'view' maybe'?) it doesn't seem to be used elsewhere.

watson/cli.py
```diff
@@ -651,7 +653,12 @@ def log(watson, current, from_, to, projects, tags, year, month, week, day,
             for frame in frames
         ))
 
-    click.echo_via_pager('\n'.join(lines))
+    if pager or (pager is None and
+                 watson.config.getboolean('options', 'log_pager',
+                                          default=True)):
```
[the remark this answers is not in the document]
MinchinWeb: I've made the removal like you suggested, but it does change the default behaviour to now bypass the pager. I've added a note to the changelog to try and make this obvious.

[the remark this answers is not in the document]
SpotlightKid: Sorry, I didn't make my self clear enough. I only meant:

```
if pager or (pager is None and
                watson.config.getboolean('options', 'log_pager', True)):
```

MinchinWeb:
Changes have been rebased onto master (and force-pushed here).

Pager selective support has been added to the `report` command. As a by-product of this, all the lines have to be gathered before they're printed, so this make make the command feel slower to some users.

The configuration option for both has been consolidated to `options.pager`.

Doc strings for both `log` and `report` have been updated. Documentation has also been updated using the script.

watson/cli.py
```diff
@@ -476,23 +483,29 @@ def report(watson, current, from_, to, projects,
...
                     datetime.timedelta(seconds=report['time'])
                 )))
             ))
 
+    if pager or (pager is None and
+                 watson.config.getboolean('options', 'pager')):
+        click.echo_via_pager('\n'.join(lines))
+    else:
+        click.echo('\n'.join(lines))
+
```
k4nar: I think we should avoid buffering all the lines when in paging mode. The way to do it is not obvious, but maybe something like that would work:
```
lines = []

if pager or (...):
    def _print(line):
        lines.append(line)
else:
    def _print(line):
        click.echo(line)

...
_print('{project} - {time}'.format(...)
...

if pager or (...):
    click.echo('\n'.join(lines)
```

watson/cli.py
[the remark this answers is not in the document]
SpotlightKid: `-v` is often used to mean `verbose` or `version`, so I don't think it's an ideal choice. How about `-l/-L` (mnemonic `less`, but it's rather arbitrary ;))?

watson/cli.py
MinchinWeb: Done in 046bf4ebcaa8fe44af4927d575f90b6399ef330d

MinchinWeb:
Lines are now printed in "real time" if printing is direct to the terminal, rather than waiting until end.

docs/user-guide/commands.md
```diff
@@ -127,6 +127,9 @@ You can also use special shortcut options for easier timespan control:
 and `--year`, `--month` and `--week` to the current year, month or week
 respectively.
 
+If you are outputting to the terminal, you can selectively enable a pager
+through the `--pager`` option.
```
SpotlightKid: Here's one backtick too many.

docs/user-guide/commands.md
```diff
@@ -174,6 +177,7 @@ Flag | Help
 `-p, --project TEXT` | Logs activity only for the given project. You can add other projects by using this option several times.
 `-T, --tag TEXT` | Logs activity only for frames containing the given tag. You can add several tags by using this option multiple times
 `-j, --json` | Format the log in JSON instead of plain text
+`-v, --pager / -V, --no-pager` | (Don't) view output through a pager.
```
SpotlightKid: Like I said, I'd rather use another letter for the short option than "v".

docs/user-guide/commands.md
```diff
@@ -329,6 +333,9 @@ You can limit the report to a project or a tag using the `--project` and
 `--tag` options. They can be specified several times each to add multiple
 projects or tags to the report.
 
+If you are outputting to the terminal, you can selectively enable a pager
+through the `--pager`` option.
```
SpotlightKid: Here's one backtick too many.

watson/cli.py
```diff
@@ -374,6 +376,9 @@ def report(watson, current, from_, to, projects,
     `--tag` options. They can be specified several times each to add multiple
     projects or tags to the report.
 
+    If you are outputting to the terminal, you can selectively enable a pager
+    through the `--pager`` option.
```
SpotlightKid: Here's the source of the extra backtick.

watson/cli.py
```diff
@@ -452,46 +457,62 @@ def report(watson, current, from_, to, projects,
 
     if format_json:
         click.echo(json.dumps(report, indent=4, sort_keys=True))
+        return
+
+    lines = []
+    # use the pager, or print directly to the terminal
+    if pager or (pager is None and
+            watson.config.getboolean('options', 'pager')):
```
SpotlightKid: Should be `watson.config.getboolean('options', 'pager', True)`, no?

watson/cli.py
```diff
@@ -452,46 +457,62 @@ def report(watson, current, from_, to, projects,
...
+        use_pager = True
+        def _print(line):
+            lines.append(line)
     else:
-        click.echo('{} -> {}\n'.format(
-            style('date', '{:ddd DD MMMM YYYY}'.format(
-                arrow.get(report['timespan']['from'])
+        use_pager = False
+        def _print(line):
```
SpotlightKid: There should be a blank line before inline function definitions (see also [Travis log](https://travis-ci.org/TailorDev/Watson/builds/315085764?utm_source=github_status&utm_medium=notification)).

watson/cli.py
SpotlightKid: Also, I'd line up this line with the second `pager` above.

watson/cli.py
```diff
@@ -452,46 +457,62 @@ def report(watson, current, from_, to, projects,
...
+            click.echo(line)
+
+    _print('{} -> {}\n'.format(
+        style('date', '{:ddd DD MMMM YYYY}'.format(
+            arrow.get(report['timespan']['from'])
+        )),
+        style('date', '{:ddd DD MMMM YYYY}'.format(
+            arrow.get(report['timespan']['to'])
+        ))
+        ))
```
SpotlightKid: Line should be on the same indentation level as `_print`.

SpotlightKid:
Regardless of this PR, I really think we should move the logging / reporting functionality into a separate command line script at some point.

docs/user-guide/commands.md
MinchinWeb: fixed!

docs/user-guide/commands.md
MinchinWeb: fixed!

watson/cli.py
MinchinWeb: fixed!

watson/cli.py
MinchinWeb: fixed!

watson/cli.py
MinchinWeb: Adding 'True' maintains the current default (at least for log) of running output through the pager.

You had me remove this earlier because it was bumping it onto a second line, but the indentation is different now, so that's no longer an issue.

Both fixed!

watson/cli.py
MinchinWeb: fixed!

MinchinWeb:
All changes as requested have been submitted.

Moving logging/report to a separate module is beyond what I set out to do here, so I'll leave it for a separate PR/refactoring effort.

watson/cli.py
SpotlightKid: I was trying to say, that you don't need the argument name, i.e `default=`, not the argument value itself.

MinchinWeb: Aaah, that makes more sense. Updated as recommended!

SpotlightKid:
@MinchinWeb: Thanks for hanging in there. Great work!

MinchinWeb:
Thanks! Glad I can do my part to make the world a little nicer!

PR #178, opened 2017-12-09
Standarize formatting of `log`'s daily header line

MinchinWeb:
Rather than making the header all one colour, this makes it so the date part is highlighted like a date, and the times are highlighted like times elsewhere.

Broken off of #176


SpotlightKid:
I rebased this on master and merged it via the command line (820e96a3606f34dba95beb84657cb228f059512f). Not sure why GH doesn't pick this up and marks the PR as merged. So closing it manually.

PR #180, opened 2017-12-09
add conditional dependicies for older versions of Python on tox

MinchinWeb:
Requirements vary slightly between different version of Python. This allows tox to deal with that.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-arrow<0.11.0
+arrow!=0.11,!=0.12
 click
 requests
--- a/tox.ini
+++ b/tox.ini
@@ -3,8 +3,11 @@ envlist = flake8,py27,py33,py34,py35,py36
 skip_missing_interpreters = True
 
 [testenv]
-deps = py
-    pytest<=3.2.5
+deps =
+    ; pytest 3.3.0 dropped support for Python 3.3
+    py33: pytest<=3.2.5
+    py27,py34,py35,py36: pytest
+    py
     mock
     pytest-datafiles
     pytest-mock


MinchinWeb:
This fixes two immediate issues:
- *pytest* dropped support for Python 3.3
- *arrow* isn't installing a required dependency on Python 2.7

An alternate way to deal with this would be move the dependencies to the `setup.py` file, and use the conditional markers provided there.

SpotlightKid:
Alas, fixing the requirements for the tests to run is not enough. `arrow` is a run-time dependency, so we need to make sure that a Python 2.7 compatible version is installed alongside Watson. We could conditionally add a max version qualifier to the `install_requires` line in `setup.py`, but that would need a restructuring of the way we declare dependencies at the moment (they are parsed form `requirements.txt`). IMHO it's easier to just put the max version restriction for every Python version by having it in `requirements.txt`. As I see it, arrow releases 0.11.0 and 012.0 are simply broken.

The change to `tox.ini` setting the `pytest` version could be merged if the other two changes are removed.

k4nar:
For Python 3.3, its support was ended [a few month ago](https://www.python.org/downloads/release/python-337/), so maybe we could remove it from tox?

I agree with @SpotlightKid for Python 2.7 & Arrow. This is very unfortunate though. We should consider removing Arrow for something more stable at some point (https://github.com/sdispater/pendulum for instance).

MinchinWeb:
I've submitted a [PR to arrow](https://github.com/crsmithdev/arrow/pull/504) to fix the issue.

Version 0.11 (of *arrow*) does appear to be broken, but version 0.12 does work, if you install from source (and not from a cached wheel generated by Python 3; *pip* shares a cache between all version installed on a machine). So this issue with *arrow* will never appear if you only have Python 2.7 on your system. To test this yourself, try install *arrow* on your system by ignoring the cache (`pip2 install arrow==0.12 --no-cache`). This is also way the issue never came up on Travis (see [here](https://travis-ci.org/MinchinWeb/Watson/jobs/314116493#L462) where the *arrow* dependency is installed as expected).

So the issue is a weird combination of *pip*, *wheel*, *arrow*, and multiple Python versions installed on the same machine.

---

It may actually be possible to maintain the `requirements.txt` fill as currently used, and use conditional dependencies, if you're willing to require *setuptools* >= 36.2.1. (See the second example in the wheel documentation on [conditional dependencies](https://wheel.readthedocs.io/en/latest/#defining-conditional-dependencies).)

---

I'm just leary of pinning a dependency to a maximum version like this because for me, *watson* is installed at a system level, and with the dependency pinned I can't upgrade it for anything else. I know that is discouraged on Linux systems, but I work on Windows and this works well, barring dependency hell, as there is to "system Python" to break.

SpotlightKid:
> It may actually be possible to maintain the requirements.txt fill as currently used, and use conditional dependencies, if you're willing to require setuptools >= 36.2.1. 

But the documentation also says:

    Specifying extras via install_requires does not yet work with pip (v9.0.1 as of this writing).

MinchinWeb:
What if I just blacklist the "broken" versions of *arrow*?

SpotlightKid:
Yes, that would work too. But until there's a fixed arrow release it makes now difference to the way `requirements.txt` is atm, so I don't see a need to change it now.

SpotlightKid:
Ok, you've convinced me, I'll merge this as it is now. Let's hope that the next `arrow` version is really fixed! ;)

MinchinWeb:
Awesome!

Here's the [arrow PR](https://github.com/crsmithdev/arrow/pull/504) to watch...

MinchinWeb:
The update to *arrow* has been merged and released as version 0.12.1. PR #210 has been created to provide that change.

PR #181, opened 2017-12-18
Update frame time when renaming projects and tags

MinchinWeb:
When renaming a project or a tag, update the "last_updated" time for the affected frames.

The other command in a similiar situation was `edit`, but it appears to already do this.


watson/cli.py
```diff
@@ -1183,7 +1184,8 @@ def rename(watson, type, old_name, new_name):
         for frame in watson.frames:
             if old_name in frame.tags:
                 watson.frames[frame.id] = frame._replace(
-                    tags=[new_name if t == old_name else t for t in frame.tags]
+                    tags=[new_name if t == old_name else t for t in frame.tags],
```
SpotlightKid: There's a flake8 line-length error here:

Maybe we could make this a bit clearer:

```
                new_tags = [new_name if tag == old_name else tag
                            for tag in frame.tags]
                watson.frames[frame.id] = frame._replace(
                    tags=new_tags,
                    updated_at=updated_at
                )
```

BTW, please make sure that you have no tox failures, including in the flake8 tests, before submitting a PR.

SpotlightKid:
Thanks for the PR, good catch!

I'm weary of adding more code again to `cli.py` for which we still don't have tests yet. I think we should put everything from line 1175 on into new methods of the `Watson` class (e.g. `rename_project` and `rename_tag`) and add tests.

MinchinWeb:
The test suite previously was spewing all sorts of errors for me, so I've fixed all the tests too (see d790bf77f0d1664bc77f9302ecb946b5aff1bc66, the issue was `dateutil` + Windows + timestamps less than 3600). There remains one test that doesn't pass --> `test_push`, where it doesn't push any frames to the server....

I"ve refactored renaming projects and tags into the `watson.py` file, and added tests.

watson/watson.py
```diff
@@ -490,3 +489,48 @@ def report(self, from_, to, current=None, projects=None, tags=None,
 
         report['time'] = total.total_seconds()
         return report
+
+    def rename_project(self, old_project, new_project):
+        """Takes a project and renames it in all affected frames."""
```
SpotlightKid: Docstring summary lines in imperative form, please.

    """Rename a project in all affected frames."""

watson/watson.py
```diff
@@ -490,3 +489,48 @@ def report(self, from_, to, current=None, projects=None, tags=None,
...
+                )
+
+        self.frames.changed = True
+        self.save()
+        click.echo('Renamed project "{}" to "{}"'
+                   .format(style('project', old_project),
+                           style('project', new_project)))
+
+    def rename_tag(self, old_tag, new_tag):
+        """Takes a tag and renames it in all affected frames."""
```
SpotlightKid: """Rename a tag in all affected frames."""

watson/watson.py
```diff
@@ -490,3 +489,48 @@ def report(self, from_, to, current=None, projects=None, tags=None,
 
         report['time'] = total.total_seconds()
         return report
+
+    def rename_project(self, old_project, new_project):
+        """Takes a project and renames it in all affected frames."""
+        if old_project not in self.projects:
+            raise click.ClickException(style(
```
SpotlightKid: `watson.py`shouldn't really depend on `click`, only `cli.py` should.  I know, it was used already in `Watson.__init__`, but that has been bugging me for a long time and we shouldn't add new code using it in this file. Definitely not using exceptions or functions interacting with the console from it.

In this particular case, raise a `ValueError` here and catch that in `cli.rename`.

watson/watson.py
```diff
@@ -490,3 +489,48 @@ def report(self, from_, to, current=None, projects=None, tags=None,
...
+        for frame in self.frames:
+            if frame.project == old_project:
+                self.frames[frame.id] = frame._replace(
+                    project=new_project,
+                    updated_at=updated_at
+                )
+
+        self.frames.changed = True
+        self.save()
+        click.echo('Renamed project "{}" to "{}"'
```
SpotlightKid: User feedback should be in `cli.py`.

watson/watson.py
```diff
@@ -490,3 +489,48 @@ def report(self, from_, to, current=None, projects=None, tags=None,
...
+                   .format(style('project', old_project),
+                           style('project', new_project)))
+
+    def rename_tag(self, old_tag, new_tag):
+        """Takes a tag and renames it in all affected frames."""
+        if old_tag not in self.tags:
+            raise click.ClickException(style(
```
SpotlightKid: Same applies as in `rename_project`.

watson/watson.py
```diff
@@ -490,3 +489,48 @@ def report(self, from_, to, current=None, projects=None, tags=None,
...
+        for frame in self.frames:
+            if old_tag in frame.tags:
+                self.frames[frame.id] = frame._replace(
+                    tags=[new_tag if t == old_tag else t for t in frame.tags],
+                    updated_at=updated_at
+                )
+
+        self.frames.changed = True
+        self.save()
+        click.echo('Renamed tag "{}" to "{}"'
```
SpotlightKid: User feedback should be in `cli.py`.

watson/watson.py
```diff
@@ -490,3 +489,48 @@ def report(self, from_, to, current=None, projects=None, tags=None,
...
+            if frame.project == old_project:
+                self.frames[frame.id] = frame._replace(
+                    project=new_project,
+                    updated_at=updated_at
+                )
+
+        self.frames.changed = True
+        self.save()
+        click.echo('Renamed project "{}" to "{}"'
+                   .format(style('project', old_project),
```
SpotlightKid: I'd personally break the (previous) line after `.format(`.

tests/test_watson.py
```diff
@@ -737,3 +737,39 @@ def test_report(watson):
     assert report['projects'][0]['name'] == 'foo'
     assert len(report['projects'][0]['tags']) == 1
     assert report['projects'][0]['tags'][0]['name'] == 'B'
+
+
+# renaming project updates frame last_updated time
+def test_rename_project_with_time(mock, watson):
+    """
+    Renaming a project should update any the "last_updated" time on any
```
SpotlightKid: update any the "last_updated"

--->

    update the "last_updated"

watson/watson.py
MinchinWeb: done

watson/watson.py
MinchinWeb: done

watson/watson.py
MinchinWeb: adjusted!

watson/watson.py
MinchinWeb: done

watson/watson.py
MinchinWeb: done

watson/watson.py
MinchinWeb: moved

tests/test_watson.py
MinchinWeb: done

MinchinWeb:
This is typical of the errors I was getting:

image

Moving to a `T_OFFSET` breaks in this case, because there is a static file `frames-with-conflict` that needs "offset" time values as well. `watson.merge_report()` would have to be refactored, as it currently expects a file.

Otherwise, the changes as suggested have been made, and a test for renaming tags has been added.

SpotlightKid:
> This is typical of the errors I was getting:

Are you in a UTC-8 timezone ?

This actually uncovers a problem with `Watson._parse_date`, which expects a unix time stamp in UTC and tries to convert it into the local timezone, which, as your example shows, fail on some systems if the UTC offset is negative. To be honest, I'm a bit astonished and shocked to realize that Watson deals with local timestamps internally, instead of  UTC ones. Anyway, we'll sort this out in another issue.

watson/watson.py
MinchinWeb: done

MinchinWeb:
I'm in UTC-7 (aka Mountain Standard Time).

MinchinWeb:
Changes made as recommended, but teach me why this second form is preferred...

SpotlightKid:
It's about different levels of abstraction.

If I use `Watson.rename_project` and it raises an exception because I passed a wrong project name, I want to get a proper error message. In `cli.rename` this error message is just passed on (we could use a differnt error message here, if the error message of the original exception is not user-friendly, but in this case it's fine).

Or do you mean the formatting of the error message?

MinchinWeb:
No, you answered my questions, and it makes sense. This way, if `watson.rename_project()` is called from elsewhere, a useful error message is available.

PR #185, opened 2018-01-07
For report and log, allow specifying a timeperiod of "all"

MinchinWeb:
`--year` was a brilliant way to get "everything", but we're in a new year now...

`-a / --all` is set to use a start date of Jan 1, 1970 (close to timestamp `0`). Alternately, I guess we could try and figure out when the first frame starts, but this seemed like it would be faster at runtime and less likely to break.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This document records all notable changes to Watson. This project adheres to
   options to only output the selected information (#166).
 * Added: the `report` and `log` commands' output can now selectively be run
   through a pager (#176).
+* Added: the `report` and `log` commands now take an `--all` option to output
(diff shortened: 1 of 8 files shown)


MinchinWeb:
Now with updated documentation!

SpotlightKid:
Thanks for the PR!

Could you please update the changelog and `watson.complete` (line 39) as well? Ideally `watson.zsh-complete` should be updated as well, but it's out-of-date and I don't know enough about Zsh's completion commands to remedy this.

MinchinWeb:
Ok, I've update the Changelog and the tab completions. I'm not working on Bash or zsh, so I don't have a way to test the tab completions, so I tried to just follow the existing pattern.

PR #187, opened 2018-01-11
Fix cli output when on tags on `stop` and `remove`

MinchinWeb:
I discovered the spacing with no tags wasn't working with the `stop` and `remove` command. This fixes that.


jmaupetit:
Thanks @MinchinWeb!

PR #188, opened 2018-01-14
Windows config is stored in Roaming, not Local

MinchinWeb:
As per the default in `click` --> http://click.pocoo.org/5/api/#click.get_app_dir

### DIFF
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -2,10 +2,10 @@
 
 ## The configuration file
 
-Watson configuration and data are stored inside your user's application folder. Depending on your system, the default path might be:
+Watson configuration and data are stored inside your user's application folder. Depending on your system, the default path is likely:
 
 * **MacOSX**: `~/Library/Application Support/watson/config`
-* **Windows**: `C:\Users\<user>\AppData\Local\watson\config`
+* **Windows**: `%appdata%\watson\config`, which usually expands to `C:\Users\<user>\AppData\Roaming\watson\config`
 * **Linux**: `~/.config/watson/config`
 
 The configuration file is a typical [python configuration (INI) file](https://docs.python.org/3.4/library/configparser.html#supported-ini-file-structure), that looks like:


SpotlightKid:
Could you elaborate a bit more why this is better?

MinchinWeb:
There are actually three folders that application data is stored in: *Roaming*, *Local*, and *LocalLow*. Actual file location have changed slightly with different version of Windows, but the these have always been hidden folders (in my experience). The variable `%appdata%` will point you to the *Roaming* folder (try typing it in either on the console or in the address bar of Explorer), and the variable `%localappdata%` will point you to the *Local* folder. In practice, there is not difference between the *Roaming* and *Local* folders for most home users. In corporate environments, if properly set up, the *Roaming* folder is copied from a central server each time you log on to a machine (and copied back to the central server when you log off) (along with a portion of the Registry), while the *Local* folder stays on the machine in front of you. The idea is that you can have the same user experience, even if you are logging on to a new machine each day. However, large roaming profiles (i.e. > 20 MB or lots of little files) can impact logon performance, as Windows won't take you to your desktop until your whole profile has been copied from the network to the local machine. *LocalLow* seems to be reserved for lower access applications. Some documentation points to Internet Explorer using this folder, but on my local machine, my password manager seems to be the only program really using the folder.

So both *Roaming* and *Local* AppData folders store the same type of information, it's just where it's stored, and whether it's automatically copied to a new machine.

As this applies to **Watson**, I think that if there is nothing in the configuration tied to a certain machine (like a filepath), then config should be stored in *Roaming* (this is the current status). If the config ever starts storing machine-specific information (like filepaths), then the config, or at least that part of it, should be stored in *Local*, because it might break moving to a new machine. If we were to move to a more Windows-native approach, perhaps the config values should be stored in the registry, and only the `frames` file stored in *AppData*.

The most important part, though, was I just want the documentation to match the facts on the ground, so I can find and manually edit the configuration, if needed.

jmaupetit:
Thank you for this in-depth, clear explanation (sorry it's been a long time since I haven't used Windows). As you stated, mentioning `%appdata%/watson` seems fine as long as it rings a bell to Windows users.

MinchinWeb:
I've changed it slightly to show the config is stored in `%appdata%`, but also noted what the full path usually is.

MinchinWeb:
@jmaupetit can we merge this? :)

jmaupetit:
Yes, thanks! Sorry for the delay...

MinchinWeb:
just glad to get it in!

PR #198, opened 2018-03-13
[WIP] add option to add aliases to commands

rndmh3ro:
I am trying to make it possible to add aliases to commands, so you can type e.g. `watson remove` or `watson delete`.

This works, however it breaks the documentation by deleting the output for:
```
 @click.argument('id')
 @click.option('-f', '--force', is_flag=True,
               help="Don't ask for confirmation.")
```

If I move these two options below `cli.command` it works. I'm not sure why.

Also the `help` is broken for `watson remove` (but not `watson delete`):

```
basti$ watson help delete
Usage: watson help [OPTIONS] ID

  Remove a frame. You can specify the frame either by id or by position (ex:
  `-1` for the last frame).

Options:
  -f, --force  Don't ask for confirmation.
  --help       Show this message and exit.
(watson) [~/git/Watson] (alias_for_commands)
```
See the missing `ID` and `-f` options:
```
basti$ watson help remove
Usage: watson help [OPTIONS]

  Remove a frame. You can specify the frame either by id or by position (ex:
  `-1` for the last frame).

Options:
  --help  Show this message and exit.
```

I guess the reason for these problems are the same, however I don't know how to fix this.

What do you think of this PR?

### DIFF
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -275,7 +275,7 @@ Flag | Help
 ## `remove`
 
 ```bash
-Usage:  watson remove [OPTIONS] ID
+Usage:  watson remove [OPTIONS]
 ```
 
 Remove a frame. You can specify the frame either by id or by position
@@ -285,7 +285,6 @@ Remove a frame. You can specify the frame either by id or by position
 
 Flag | Help
 -----|-----
-`-f, --force` | Don't ask for confirmation.
 `--help` | Show this message and exit.
 
 ## `rename`
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -44,6 +44,27 @@ class WatsonCliError(click.ClickException):
         return style('error', self.message)
 
 
+class CustomMultiCommand(click.Group):
+    def command(self, *args, **kwargs):
+        """Behaves the same as `click.Group.command()` except if passed
+        a list of names, all after the first will be aliases for the first.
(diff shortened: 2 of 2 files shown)


k4nar:
There is an example in Click's documentation: http://click.pocoo.org/5/advanced/#command-aliases . Maybe it would solve those issues?

However, I wonder if it's a good idea to introduce aliases. I think we should rather aim to be consistent in our naming.

willdurand:
> However, I wonder if it's a good idea to introduce aliases. I think we should rather aim to be consistent in our naming.

I second this to be honest.

rndmh3ro:
I'll submit to your reasoning! :)

raphaelfournier:
I may be rather new to watson, but as a heavy Git user (like the folks at TailorDev seems to be), I would welcome warmly an option to use `watson st` for `watson status`, `watson be` for `watson start`, etc., Like `git ci` is a widely popular alias for `git commit`. As a user, not a dev, I don't exactly get what you mean by "aim to be consistent in our naming".

So consider this as a feature request.

jmaupetit:
Hey @raphaelfournier :wave: Thanks for sharing your thoughts! IMO allowing users to define their own aliases is not in contradiction with being consistent in our naming. And, as a daily git shortcuts user (:smiling_imp:), I would also like to see this feature added to Watson.

PR #202, opened 2018-03-21
Add frame functionality, posthumously recording a task

braincow:
At times I found myself starting a project, stopping and afterwards editing it to input in a task I forgot to track live earlier. With this functionality it is now possible to fill in these tasks afterwards with simple command line magic instead.

Includes  BASH completion and pytest unit tests. ZSH completion is missing as I do not posses that particular shell fu.


watson/cli.py
```diff
@@ -773,6 +772,45 @@ def frames(watson):
...
+    # Find all the tags starting by a '+', even if there are spaces in them,
+    # then strip each tag and filter out the empty ones
+    tags = list(filter(None, map(operator.methodcaller('strip'), (
+        # We concatenate the word with the '+' to the following words
+        # not starting with a '+'
+        w[1:] + ' ' + ' '.join(itertools.takewhile(
+            lambda s: not s.startswith('+'), args[i + 1:]
+        ))
+        for i, w in enumerate(args) if w.startswith('+')
+    ))))  # pile of pancakes !
```
k4nar: we should write a (unit tested) utils for that :)

braincow: I thought for a minute of separating that into its own function to avoid replicating code, but then I got lazy. But you are correct; separating that tag parsing into its own function and writing an unit test for that is a good idea.

beberlei:
I am very much a late recording kind of time tracker, so i would appreciate this functionality very much.

braincow:
I do not want to sound way too eager beaver but, I am wondering, what is the ETA for next Watson release and if this merge could be part of it? I have been working with a companion tool for Watson for some time now and it requires this modification to exist in upstream first to be installable. It is called Flubber and it is a GTK+ front-end for Watson. Feel free to check it out at https://github.com/braincow/flubber

BuddhaOhneHals:
I would love this feature. What needs to be done to get this one merged? Perhaps I can help.

jmaupetit:
@BuddhaOhneHals :heart: can you help with missing zsh-completions?

BuddhaOhneHals:
@jmaupetit @braincow 
I separated the parsing of the tags and added tests: https://github.com/braincow/Watson/pull/1

Additionally I added zsh completion. I use zsh, but it's the first time I wrote completions for it. The changes seem to work just fine but a second look from someone with more experience could be helpful.

jmaupetit:
:thinking: looks like we are in trouble with python 3.4 I've restarted the job. Let see how it goes.

PR #209, opened 2018-06-04
[WIP] Add the capability to remove a given project and all related frames

### LINKED ISSUE #192: Possibility to remove the project
It would be good to delete the entire project.

jnsebgosselin:
Fixes #192

- [x] Add a method to `Watson` class to delete project by name
- [ ] Add the functionality to the CLI
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/watson/watson.py
+++ b/watson/watson.py
@@ -506,6 +506,17 @@ class Watson(object):
         self.frames.changed = True
         self.save()
 
+    def delete_project(self, project):
+        """Delete the project and all related frames."""
+        if project not in self.projects:
+            raise ValueError('Project "%s" does not exist' % project)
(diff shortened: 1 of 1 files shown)


k4nar:
Thanks for the PR!

As it's a dangerous operation, I think we should ask the user for confirmation. Here is a pattern for doing that with Click: http://click.pocoo.org/5/options/#yes-parameters

Ideally, we should also tell the user how many frames this will delete.

jnsebgosselin:
@k4nar for the pointer.

I agree. I'll do that when I implement this feature in the CLI.

napstercc (approved):

PR #210, opened 2018-06-05
Allow arrow v0.12.1

MinchinWeb:
arrow versions 0.11.0 and 0.12.0 are both broken, but 0.12.1 works. This builds off of #175 and #180.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-arrow!=0.11,!=0.12
+arrow!=0.11,!=0.12.0
 click
 requests


k4nar:
I think we need to drop the Python 3.3 job as well. This is an unsupported version of Python, and apparently it causes issues with Flake8.

MinchinWeb:
That's a decision I can support!

(The error on Travis appears to actually be with *wheel*, which dropped support for Python 3.3. So then Travis can't get the tox environment set up correctly. "flake8" is just the name of the first environment tox tries to set up.)

k4nar:
> (The error on Travis appears to actually be with wheel, which dropped support for Python 3.3. So then Travis can't get the tox environment set up correctly. "flake8" is just the name of the first environment tox tries to set up.)

Ok sorry I misread the error :) .

MinchinWeb:
Dropped the classifier from *setup.py*

PR #225, opened 2018-09-21
Negate option for projects and tags options in log/report commands

arnaudcordier:
Hi,

One use case I usually need in a log or a report is to show all projects but one or two, or frames with all tags but one.

So here is --exclude-tags and --exclude-projects flag option that exclude given tags or projects by --tag or --project options in log and report command.

In order to be able to type:  
`watson log -Gd -p coding -T personal --exclude-tags`  
to have log of code activity of the day, but not the personal ones.

### DIFF
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -194,6 +194,7 @@ def test_frames_with_empty_given_state(config_dir, mock):
     mock.patch('%s.open' % builtins, mock.mock_open(read_data=content))
     assert len(watson.frames) == 0
 
+
 def test_frames_filter(watson):
     samples = (
         ('foo', ('A')),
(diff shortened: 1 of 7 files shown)


tests/test_watson.py
```diff
@@ -195,6 +195,75 @@ def test_frames_with_empty_given_state(config_dir, mock):
...
+    )
+    for name, tags in samples:
+        watson.frames.add(name, 4000, 4000, tags)
+
+    foo_projects = watson.frames.filter(projects=('foo'))
+    len = 0
+    for frame in foo_projects:
+        assert frame.project == 'foo'
+        len += 1
+    assert len == 2
```
k4nar: I would find the following more readable:
```
assert all(frame.project == 'foo' for frame in foo_projects)
assert len(foo_projects) == 2
```

This is true for all the other cases of course :) .

jmaupetit:
@arnaudcordier is there any chance that you pursue this work or not? The second is perfectly fine if you don't have time for this.

In fact, your work on the `report` command has already been merged (see #288), but not for the `log` command.

Anyway, thank you again for this :pray:

arnaudcordier:
Hi, sorry I didn't really follow this issue.
Anyway I reworked on it and made a new PR using work that had been done on the project command : #317

PR #227, opened 2018-10-11
Fix zsh completion for project and tag names containing spaces

jbernhard:
This fixes a bug where project and tag names containing spaces were split on the spaces and completed as separate entities.  For example, given the project name "one two three", zsh would complete "one", "two", and "three" separately.  Typing

```
watson start o<tab>
```

would expand to

```
watson start one
```

but now correctly expands to

```
watson start one\ two\ three
```

### DIFF
--- a/watson.zsh-completion
+++ b/watson.zsh-completion
@@ -42,7 +42,7 @@ _watson_get_projects() {
   if ( [[ ${+_watson_project_list} -eq 0 ]] || _cache_invalid $cacheid ) \
             && ! _retrieve_cache $cacheid;
   then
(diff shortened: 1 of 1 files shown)


k4nar:
Great :+1: !

Do you know if there is a way to have it between quotes instead?

jbernhard:
If you type a quote first, it will complete without the backslashes and close the quote, e.g.

```
watson start 'o<tab>  ->  watson start 'one two three'
```

I don't know of a way to make it always use quotes, but zsh has so many options, maybe there is something...

k4nar:
Ok, sounds fair enough :) . Thank you for your contribution!

Issue #234, opened 2018-11-11
Start new frame seamlessly to old frame


MaxG87:
Sometimes I want to log time for several frames already finished. An exemplaric use case is given in 
https://github.com/TailorDev/Watson/issues/233.

It would be great if there was a way to start a new frame seamlessly to the old one. Then, at the end of several short meetings, one could issue

    watson start Project_1 +Meetings
    #Go chatting to several colleagues
    watson stop --at 13:37
    watson start Project_2 +Meetings --seamless
    watson stop --at 13:51

This PR is closely related to https://github.com/TailorDev/Watson/issues/75, https://github.com/TailorDev/Watson/issues/233 and https://github.com/TailorDev/Watson/issues/28.

Thank you very much.

MaxG87:
I have nothing prepared, but I will give it a try.

MaxG87:
As you can see in https://github.com/TailorDev/Watson/pull/235, I managed to dive into the codebase. It will take me a week or two. I hope that this is ok.

jmaupetit:
Of course it's ok! You owe us nothing. Take your time.

MaxG87:
Shouldn't this issue be closed now?

Issue #233, opened 2018-11-11
Stop frame at given time


MaxG87:
Hi and thank you for that awesome program.

Sometimes I want to log time for several frames already finished. The typical use case would be that you have a short conversation/ad-hoc meeting with colleague A on Project 1, then you switch on Project 2 and then you visit colleague B to talk about Project 1 again. One cannot log times for that instantly but has a rough idea about the times. This kind of use case is not too seldom, at least for me.

Currently, this kind of use case requires extensive use of `watson edit`, which is quite burdensome. It would be great to have a `--at` option, so that one can issue

    watson start Meetings +Project_1
    #Go chatting to several colleagues
    watson stop --at 13:37

This request is closely related to https://github.com/TailorDev/Watson/issues/75.

I will request corresponding flags on `watson start` in a seperate PR.

loonies:
What about `watson stop --at 12:34` to match `watson start --at` ?

acidjunk:
Very nice feature (handy for start, stop and restart)

varac:
Isn't this already implemented by aa901567c5aa6129ff6dae799eddbfb0be06cb65 ?

k4nar:
Indeed :+1: .

PR #236, opened 2018-11-25
Provide support to  add multiple passed time trackings

MaxG87:
Merging this PR will add the functionality needed to support the workflow described in 
https://github.com/TailorDev/Watson/issues/233 and https://github.com/TailorDev/Watson/issues/234.

There are some caveats I want to mention. Please give me a hint how to resolve them or feel free to alter my commits.

  * For '--seamless', I use the last frame instead of the most recently stopped. This might cause problems after using `watson edit`. I wanted to avoid looping through the list of frames for efficiency reasons.
  * Using '-s' as short option for '--seamless' will block adding the '--stop/--no-stop' option from `watson restart` to `watson start` too.
  * The '--at' option uses the current wall clock day to complete the given time to a parseable time stamp string. This might cause issues around midnight or for very long lasting sessions.

Please let me know if should improve or add something.

### DIFF
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -642,7 +642,7 @@ Example:
 
 
     $ watson stop --at 13:37
-    Stopping project apollo11, started a minute ago, at 30 minutes ago. (id: e9ccd52) # noqa: E501
+    Stopping project apollo11, started a minute ago, at 30 minutes ago. (id: e9ccd52)
 
 ### Options
 
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -577,16 +577,20 @@ If there is already a running project and the configuration option
 `options.stop_on_start` is set to a true value (`1`, `on`, `true` or
 `yes`), it is stopped before the new project is started.
 
+If the '--no-gap' flag is given, the start time of the new project is set
+to the stop time of the most recently stopped project.
+
 Example:
 
 
-    $ watson start apollo11 +module +brakes
+    $ watson start apollo11 +module +brakes --no-gap
     Starting project apollo11 [module, brakes] at 16:34
 
 ### Options
 
 Flag | Help
 -----|-----
+`-g, --gap / -G, --no-gap` | (Don't) leave gap between end time of previous project and start time of the current.
 `--help` | Show this message and exit.
 
 ## `status`
@@ -630,16 +634,21 @@ Usage:  watson stop [OPTIONS]
 
 Stop monitoring time for the current project.
 
+If '--at' option is given, the provided stopping time is used. The
+specified time must be after the begin of the to be ended frame and must
+not be in the future.
+
 Example:
 
 
-    $ watson stop
-    Stopping project apollo11, started a minute ago. (id: e7ccd52)
+    $ watson stop --at 13:37
+    Stopping project apollo11, started a minute ago, at 30 minutes ago. (id: e9ccd52) # noqa: E501
 
 ### Options
 
 Flag | Help
 -----|-----
+`--at TIME` | Stop frame at this time. Must be in (YYYY-MM-DDT)?HH:MM(:SS)? format.
 `--help` | Show this message and exit.
 
 ## `sync`
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -642,7 +642,7 @@ Example:
 
 
     $ watson stop --at 13:37
-    Stopping project apollo11, started a minute ago, at 30 minutes ago. (id: e9ccd52)
+    Stopping project apollo11, started an hour ago and stopped 30 minutes ago. (id: e9ccd52)
 
 ### Options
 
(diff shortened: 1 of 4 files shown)


MaxG87:
Hi,

I just wanted to ask what the status is of this PR.

I noticed that there are two checks failing, but theses are due to too long lines. However, the too long lines are caused by a combination of indentation and text length and I cannot see a way to improve this without harming the overall style more than ~85 character lines are doing. I would appreciate feedback on this very much.

Sincerely

jmaupetit:
Hi @MaxG87. Concerning the two flake8 long line issues:

1. You can safely ignore the docstring with a `#noqa` flag
2. You can format long strings with the tuple trick, _i.e._:

```python
errmsg = (
    'Could not parse time. '
    'Please specify in HH:MM(:SS)? format.'
)
```

When checking the CI logs, I can see that some tests are failing too. Can you check these?

MaxG87:
Sorry, I was confused by the outputs of flake8 and thought it's only the line length.

jmaupetit:
Don't be sorry! Can you update the documentation _via_ a `make docs`?

See http://tailordev.github.io/Watson/contributing/hack/

loonies:
Alternatives to `--seamless`: `-f, --follow`, `-c, --continue`

MaxG87:
I improved the PR a bit with things that bothered me while using it locally.

@jmaupetit I also updated the documentation, but the only effect I see is that it removes the luna related documentation. I find that weird, so I would welcome if you could double check on this.

@loonies @jmaupetit I am in favour of a better name while I do not like `--follow` or `--continue` so much. What about `--ensuing`?

loonies:
I don't have any better suggestion

andrewmeyer:
`-n --nogap`?

MaxG87:
I like `-n --nogap`. Could a member, e.g. @jmaupetit or @k4nar decide which to take, please?

k4nar:
`--nogap` is probably more explicit indeed. Maybe we could take it further with `--from-last` ?

loonies:
`--nogap` doesn't sound bad

k4nar:
@MaxG87 : I think you need to run `make docs` again.

MaxG87:
Thank you for reminding me. Everything should be fine now.

docs/user-guide/commands.md
```diff
@@ -630,16 +634,21 @@ Usage:  watson stop [OPTIONS]
...
+specified time must be after the begin of the to be ended frame and must
+not be in the future.
+
 Example:
 
 
-    $ watson stop
-    Stopping project apollo11, started a minute ago. (id: e7ccd52)
+    $ watson stop --at 13:37
+    Stopping project apollo11, started a minute ago, at 30 minutes ago. (id: e9ccd52) # noqa: E501
```
jmaupetit: I think we can manually remove the `# noqa: E501` comment :wink:

watson/cli.py
```diff
@@ -68,7 +69,33 @@ def convert(self, value, param, ctx):
...
 
+class TimeParamType(click.ParamType):
+    name = 'time'
+
+    def convert(self, value, param, ctx):
+        if isinstance(value, arrow.Arrow):
+            return value
+
+        date_pattern = '\d{4}-\d\d-\d\d'  # noqa: W605
+        time_pattern = '\d\d:\d\d(:\d\d)?'  # noqa: W605
```
jmaupetit: Have you seen this? 
https://lintlyci.github.io/Flake8Rules/rules/W605.html

docs/user-guide/commands.md
MaxG87: Sorry, I don't get that. If I remove it, flke8 complains.

watson/cli.py
MaxG87: You are right. It should be fixed now.

docs/user-guide/commands.md
k4nar: I think @jmaupetit meant to remove it in the `.md` document, which isn't concerned by flake8 :) .

jmaupetit: Indeed :wink:

MaxG87:
I just wanted to ask what the status is of this PR. For me it seems as if all issues were resolved. Is that correct?

MaxG87:
I applied the requested changes and rebuilt the docs. However, I found myself unable to squash the commits in. I would have had to squash that to a commit so far back in history that I would have invalidated plenty of the more recent commits. Fixing this would have resulted in history engineering and I would consider that to be bad practice.

If you now find the history of the feature to be to complex, you might like the single commit version I created (https://github.com/TailorDev/Watson/pull/254). I simply squashed all commits into on big commit, leading to what some might consider to be a nicer history.

However, I would very much prefer to use the honest history, featuring many small and atomic commits but revealing some of the wrong paths that were taken.

loonies:
I just started using 1.7 and this is great! Thanks!

PR #240, opened 2018-12-09
Add option to specify start of the week

pedrorivotti:
This PR adds a configurable option, `options.week_start`, that allows specifying the start day of a week for reporting and logging purposes. 

This is useful for those who (like me) need to report weekly time based on a starting day other than Monday. Sunday would be the other common starting day, but I've added the option to specify any day of the week, for completeness. 

I've included tests, and a short entry in the docs.

Open to suggestions, of course!

### DIFF
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -14,9 +14,9 @@ import click
 
 from . import watson as _watson
 from .frames import Frame
-from .utils import (format_timedelta, get_frame_from_argument,
-                    get_start_time_for_period, options, safe_save,
-                    sorted_groupby, style, parse_tags)
+from .utils import (apply_weekday_offset, format_timedelta,
+                    get_frame_from_argument, get_start_time_for_period,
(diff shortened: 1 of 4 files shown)


jmaupetit:
Hi @pedrorivotti ! Thanks for this useful feature :pray: We'll review it quickly.

watson/cli.py
```diff
@@ -57,6 +57,13 @@ def convert(self, value, param, ctx):
             # expected by the user, so that midnight is midnight in the local
             # timezone, not in UTC. Cf issue #16.
             date.tzinfo = tz.tzlocal()
+
+            # Add an offset to match the week beginning specified in the
+            # configuration
+            if param.name == "week":
+                week_start = ctx.obj.config.get("options", "week_start")
```
k4nar: Shouldn't we set a default as well?

pedrorivotti: Thanks, @k4nar , well spotted!

k4nar:
Thanks!

PR #241, opened 2018-12-10
support py37

ThomasWaldmann:


.travis.yml
```diff
@@ -1,10 +1,34 @@
...
+matrix:
+    include:
+        - python: 3.4
+          os: linux
+          dist: trusty
+          env: TOXENV=flake8
+        - python: 2.7
+          os: linux
+          dist: trusty
+          env: TOXENV=py34
```
k4nar: shouldn't that be TOXENV=py27 ?

ThomasWaldmann: good spot. c&p error, fixed only one place...

i'll update the PR ...

PR #253, opened 2019-02-09
Cached projects/tags for fish to fix #250

### LINKED ISSUE #250: Slow fish completions
## Context (Environment)
- Terminal: rxvt-unicode-pixbuf 9.22-2
- Shell: fish-3.0.0-1
- Version: watson 1.6.0-1
- Completion: latest version from https://github.com/TailorDev/Watson/commit/f1a2c94c1878a6f0cace8bdd509628990e319c65

## Expected Behavior
Completions as fast as `zsh`. Here I am using `watson add` (`wa` is my alias) in `zsh`:

The projects almost come up instantly, and the `+` for the tags right after comes almost immediately.

## Current Behavior
Pretty unexpectedly slow completions in `fish`.

The projects take a few seconds to come up, and the `+` for the tags right takes bizarrely long to show up, even though it's not actually displaying any tags yet.

## Proposal
Could the speed of this be improved in any way, or is it just a problem on my own machine?

Kevin-Mok:
Formalized my workaround mentioned in [this comment](https://github.com/TailorDev/Watson/issues/250#issuecomment-462029698). Cache is invalidated when the frames file is newer than the cache files, similar to how the `zsh` completions work.


andrewmeyer:
from what I can tell this greatly increases the completion speed for projects and tags, thanks for figuring this one out!  I can provide the fuzzer I used if anyone needs it.

```
chimera  ~  watson tags | wc -l
7
chimera  ~  watson projects | wc -l
882
chimera  ~  watson frames | wc -l
5282
```

jmaupetit:
Thanks @Kevin-Mok & @andrewmeyer :pray: If you can rebase your work, I'll merge it asap @Kevin-Mok

Kevin-Mok:
This still ended up being too slow for my liking, so what I ended up doing was manually updating the cache files and just `cat`'ing them directly for my completions. The problem is that the frames file will always be newer when you start or add another frame, so the cache files will always have to be regenerated.

Would it be possible to keep and update a cache file (similar to how this works now) in the the new `confirm_project` and `confirm_tags` functions? This would fix the problem of having to constantly regenerate the cache files when there might not even be any new project or frame being added. These files could also then be used for the `bash` and `zsh` completions which suffer the same problem.

PR #254, opened 2019-02-09
Add `start --no-gap` and `stop --at`

MaxG87:
This commit adds the functionality needed to catch up on time tracking.

Take the scenario where you are engaged in several ad hoc discussions at
the desks of your colleagues, seamlessly transitioning into each other.
While you manage to remember when each discussion start, you still need
to enter those into Watson.

This commit gives you the tools to do so conveniently. You can use

    watson stop --at

to specify when a event stopped and

    watson start --no-gap

to instruct Watson to start the new event seamlessly to the one most
recently stopped. By this to catch up on your time tracking all you need
to do is to cycle through these two commands.

### DIFF
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -577,16 +577,20 @@ If there is already a running project and the configuration option
 `options.stop_on_start` is set to a true value (`1`, `on`, `true` or
 `yes`), it is stopped before the new project is started.
 
+If the '--no-gap' flag is given, the start time of the new project is set
+to the stop time of the most recently stopped project.
+
 Example:
 
 
-    $ watson start apollo11 +module +brakes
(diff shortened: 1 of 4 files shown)


MaxG87:
In case you do not like the git history of https://github.com/TailorDev/Watson/pull/236, I created this branch to serve as a single commit version of the feature. It can be cherry-picked/fast-forward merged, leading to a linear history. Some prefer those.

Please note that I am much more in favour of the actual development history of https://github.com/TailorDev/Watson/pull/236 as it offers many atomic commits and some insights in the development process.

jmaupetit:
I've rebased the work in #236 and merged it. For external contributions, I prefer having a single commit feature or fix and keep a linear comprehensive history. But I understand your point. Thanks again for your work!

PR #266, opened 2019-04-04
Add bash completion for stop command

MaxG87:


MaxG87:
I noticed other, similar feature requests. I will provide a new PR.

PR #268, opened 2019-04-05
Provide Bash completion for `watson aggregate`

MaxG87:
This commit provides complete Bash completion for `watson aggregate`.
The completion supports:

  * completing the 'aggregate' command itself
  * suggesting projects or tags after the respective options
  * suggesting available long options when at least a dash was typed
  * doing nothing otherwise (i.e. not suggesting long options for valid
    command lines)

### DIFF
--- a/watson.completion
+++ b/watson.completion
@@ -5,7 +5,7 @@
 _watson_complete () {
   local cur cmd commands frames projects tags
 
-  commands='add cancel config edit frames help log merge projects remove rename report restart start status stop sync tags'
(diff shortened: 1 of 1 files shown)


MaxG87:
I am not too satisfied with the structure of the completion code. For my taste, there is too many small scale code duplication and a bit too much branching. If someone knows how to do better, I would love to learn!

MaxG87:
This PR should be obsolete as soon as #276 got merged. Therefore I close it.

PR #243, opened 2018-12-15
Add some information in the documentation for beginners.

raphaelfournier:
Questions raised from the previous version and answered by the new paragraphs:
- Which files should one edit after setting up the environment?
- How to copy some real frames from a working watson install towards the
development one?
- How to set up the environment variable WATSON_DIR?
- How to avoid the incompatibility between mkdocs 0.14 and Python 3.7?

Most information was given by @jmaupetit in https://github.com/TailorDev/Watson/issues/238.

Fix #238

Issue #238, opened 2018-11-30
Add more precise directions on how to code/test in the Contributing documentation


raphaelfournier:
I am willing to contribute to Watson. I followed the instructions currently in the http://tailordev.github.io/Watson/contributing/hack page, everything is OK (forking, virtualenv, branch creation, tox).
However, I am missing some directions on how to work at the `Now you can make your changes locally.` step. 

I have a watson installed on my machine, and the development version. What files from my development repo should I edit? I guess they are in the `watson` subfolder. I see a  `build/lib/watson` subfolder, I guess I have to use `make` or something so that files from `watson/` are moved to `build/lib/watson`: which command(s) do(es) that?

Also, once I will have made some changes (and *built* them?), how do I test the new watson behavior/command?  

For instance, I get this error:
```
$ python watson/watson.py
Traceback (most recent call last):
  File "watson/watson.py", line 18, in <module>
    from .config import ConfigParser
ModuleNotFoundError: No module named '__main__.config'; '__main__' is not a package
```
I would also like to use some frames from my production watson, to test some features, but would not like to mess them up. How may I copy some of them so that they are used by the development version?

jmaupetit:
Hi @raphaelfournier!

You are right: the documentation needs a revamp to easily start working on the project. I'll try to answer your questions:

> What files from my development repo should I edit? I guess they are in the `watson` subfolder.

Yes, core files from the project are under the `watson/` directory.

> I see a build/lib/watson subfolder, I guess I have to use make or something so that files from watson/ are moved to build/lib/watson: which command(s) do(es) that?

The `build` directory is created while trying to build (or install) Watson with the setup script, _e.g._:

```bash
$ python setup.py build
# or
$ python setup.py install
```
Files from the `build` directory are temporary and their modification is not intended in the development workflow. You should focus on the `watson` directory.

> Also, once I will have made some changes (and built them?), how do I test the new watson behavior/command?

You should work in a python virtualenv; once activated (and Watson installed in this virtualenv) the `watson` command should point to this development installation:

```bash
# Create a virtualenv
$ python3 -m venv venv
# Activate this virtualenv
$ source venv/bin/activate
# Install Watson's dependencies in this virtualenv
(venv) $ pip install -r requirements-dev.txt
# Install Watson in development mode
(venv) $ python setup.py develop
# Use freshly installed development version of Watson
(venv) $ venv/bin/watson --help
```
Remarks:

[...]

PR #273, opened 2019-04-18
Modify WATSON_DIR on virtualenv activation

davidag:
I'm planning to contribute to watson, so I'm going through the repo and docs, trying to hack a bit in my local environment.

After reading the hack documentation and the issue #238 I think there is a potential risk of messing with your real watson data, which is not desired at all (although yes, I have backups :)

In this PR, I just make sure that `WATSON_DIR` will always be equal to `"$(watson_dir)/data"` when activating the virtualenv.

I can think of a couple of things against this PR: 
* The "data" directory name is only based on the comments of #238 but people could be using a different name. If this would be the case, it doesn't affect them because they're doing an export after activating the environment anyway.
* No documentation. I think the hack doc needs some work, but first PR #243 should be merged or rejected. After that, it would be nice to speak of the Makefile on that document, which would include a mention to this PR change. Also, adding Python3 support to the aforementioned Makefile could be interesting.

Any comments are welcomed!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,11 +6,13 @@ PIP ?= pip
 VENV = virtualenv
 VENV_ARGS = -p $(PYTHON)
 VENV_DIR = $(CURDIR)/.venv
+VENV_WATSON_DIR = $(CURDIR)/data
 
 all: install
 
 $(VENV_DIR): requirements-dev.txt
 	$(VENV) $(VENV_ARGS) "$(VENV_DIR)"
+	echo "export WATSON_DIR=\"$(VENV_WATSON_DIR)\"" >> "$(VENV_DIR)"/bin/activate
 	"$(VENV_DIR)"/bin/pip install -U setuptools wheel pip
 	"$(VENV_DIR)"/bin/pip install -Ur $<
 


Makefile
```diff
@@ -6,11 +6,13 @@ PIP ?= pip
 VENV = virtualenv
 VENV_ARGS = -p $(PYTHON)
 VENV_DIR = $(CURDIR)/.venv
+VENV_DATA_DIR = $(CURDIR)/data
```
jmaupetit: Even if it will be loaded with the virtualenv, I think we can call it `WATSON_DIR` to be more explicit:
```suggestion
WATSON_DIR = $(CURDIR)/data
```

Makefile
```diff
@@ -6,11 +6,13 @@ PIP ?= pip
...
 
 all: install
 
 $(VENV_DIR): requirements-dev.txt
 	$(VENV) $(VENV_ARGS) "$(VENV_DIR)"
+	echo "export WATSON_DIR=\"$(VENV_DATA_DIR)\"" >> "$(VENV_DIR)"/bin/activate
```
jmaupetit:
```suggestion
	echo "export WATSON_DIR=\"$(WATSON_DIR)\"" >> "$(VENV_DIR)"/bin/activate
```

davidag: @jmaupetit what about `VENV_WATSON_DIR`? I find having two `WATSON_DIR` variables in two languages (shell & makefile) in the same expression a bit confusing. Besides, prefixing with `VENV` seems quite consistent and meaningful in that context. What do you think? If I didn't convince you I'll go with `WATSON_DIR` , no problem :)

jmaupetit: I find this more consistent, but `VENV_WATSON_DIR` is ok to me (at least more relevant than `VENV_DATA_DIR` :wink:).

davidag:
I've modified the variable name and rebased onto master. I hope there's nothing missing or wrong, but if that would be the case, just let me know and I'll fix it. Thank you!

PR #218, opened 2018-08-24
 start,add and edit can confirm new project or tag 

braincow:
In issue https://github.com/TailorDev/Watson/issues/191 possibility to configure Watson to warn about adding projects and/or tags that do not exist yet was requested.

This pull requests now implements this feature. By adding --confirm-new-project or --confirm-new-tag to start, add or edit commands this feature is enabled. By default --no-confirm-new-project and --no-confirm-new-tag are in effect. This feature can also be enabled by adding confirm_new_project = true and/or confirm_new_tag = true to Watson configuration file under 'options' section.

PR #275, opened 2019-04-20
Warn new project

medwig:
A re-implementation of this PR: https://github.com/TailorDev/Watson/pull/218

Quoting:
```
In issue #191 possibility to configure Watson to warn about adding projects and/or tags that do not exist yet was requested.

This pull requests now implements this feature. By adding --confirm-new-project or --confirm-new-tag to start, add or edit commands this feature is enabled. By default --no-confirm-new-project and --no-confirm-new-tag are in effect. This feature can also be enabled by adding confirm_new_project = true and/or confirm_new_tag = true to Watson configuration file under 'options' section.
```

With the following changes to that PR:
Adds the missing unit tests.
Refactors to keep watson object separate from utils module.
Simplifies the confirm_ methods by letting click raise the Abort exception automatically if user rejects. 
Removes --no-confirm* cli options, as the default=False arg makes those options redundant. 
Update the docs

Closes #218

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for first day of the week configuration in reports and logs (#240)
 - Python 3.7 support (#241)
 - Add `start --no-gap` and `stop --at` options (#254)
-- `--confirm-new-project` and `--confirm-new-tag` options to `start`, `add`, `edit` commands (#275)
+- Add `--confirm-new-project` and `--confirm-new-tag` options to `start`, `add` and `edit` commands (#275)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for first day of the week configuration in reports and logs (#240)
 - Python 3.7 support (#241)
 - Add `start --no-gap` and `stop --at` options (#254)
+- `--confirm-new-project` and `--confirm-new-tag` options to `start`, `add`, `edit` commands (#275)
 
 ### Changed
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -12,10 +12,14 @@ except ImportError:
     from io import StringIO
 
 import pytest
+from unittest.mock import patch
+from click.exceptions import Abort
 from dateutil.tz import tzutc
 
+import watson
 from watson.utils import (apply_weekday_offset, get_start_time_for_period,
-                          make_json_writer, safe_save, parse_tags, PY2)
+                          make_json_writer, safe_save, parse_tags, PY2,
+                          confirm_project, confirm_tags)
 from . import mock_datetime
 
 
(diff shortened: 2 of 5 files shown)


tests/test_utils.py
```diff
@@ -174,3 +179,45 @@ def failing_writer(f):
...
+
+
+@patch('click.confirm', return_value=True)
+def test_confirm_project_accept_returns_true(confirm):
+    project = 'baz'
+    projects = ['foo', 'bar']
+    assert confirm_project(project, projects)
+
+
+@patch('watson.utils.click.confirm', side_effect=Abort)
```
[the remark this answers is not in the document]
medwig: Good point! This was just my inexperience with patching - I was under the impression the patched needed to be made at the level of the imported module itself.

medwig:
Thanks for the prompt feedback David! I pushed a commit with all the requested changes.

I took the liberty of re-formatting the `utils` import statements to be one per line, as I find this makes future commits that affect imports much cleaner. Let me know if this is goes against the intended style and I'll switch it back to the former line-break format.

watson/cli.py
```diff
@@ -997,8 +1020,12 @@ def frames(watson):
               help="Date and time of start of tracked activity")
 @click.option('-t', '--to', required=True, type=Date,
               help="Date and time of end of tracked activity")
+@click.option('--confirm-new-project', is_flag=True, default=False,
+              help="Confirm addition of new project.")
+@click.option('--confirm-new-tag', is_flag=True, default=False,
```
jmaupetit: For those two new options, can we:

1. default to `True` (and thus update the long option name)
2. add a short option?

braincow: 1) If I recall what I originally did (sorry, this feature has been dormant for some time) if user wants these as default they can set it in the config.
2) Watson command line is starting to run out short option names, at least in such a way that they are logically comparable to the long one.

medwig: @jmaupetit  - you would prefer these confirmation steps become the default configuration/behaviour of watson? Doesn't that seem like a major version update kind of change? (has the potential to break scripting that people have built around watson that isn't set up to handle a user input step)

jmaupetit: @medwig you are right, this is a BC and we should avoid it. I should have thought more before answering this morning. So. Your implementation, is the desired behavior. To me it should be the default, so, if it's configurable as @braincow says, it sounds perfect to me.

Even if we are running out of relevant short option names, I still would prefer having one that is not that obvious than none.

medwig:
Agreed! I'll add the short option names

jmaupetit:
@medwig sorry to bother you one last time: can you add this feature to the changelog please? Thx!

medwig:
> @medwig sorry to bother you one last time: can you add this feature to the changelog please? Thx!

No problem! Thanks for your patience in pointing out the things I missed :)

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for first day of the week configuration in reports and logs (#240)
 - Python 3.7 support (#241)
 - Add `start --no-gap` and `stop --at` options (#254)
+- `--confirm-new-project` and `--confirm-new-tag` options to `start`, `add`, `edit` commands (#275)
```
jmaupetit:
```suggestion
- Add `--confirm-new-project` and `--confirm-new-tag` options to `start`, `add` and `edit` commands (#275)
```

jmaupetit:
> No problem! Thanks for your patience in pointing out the things I missed :)

This point not documented yet. Same on me.

Thank you for your contribution :pray: after one last minor change, we will be ready to merge this.

Issue #248, opened 2019-01-29
Aggregate doesn't seem to work with frames that cross day boudaries


KennethNielsen:
Hi everyone

I just noticed the completely awesome aggregate feature in the development version (implemented by @brandomr and thanks!) and I was very thrilled because I really needed that while migrating tasks from another tool.

While migrating the tasks, I noticed that the command fails to show frames that cross day boundaries. I recognize that this is a rare usecase, but none the less I feel that it would be worth fixing.

I can help with it if necessary, but if @brandomr want's to continue the work on this command I completely understand.

brandomr:
@KennethNielsen that is interesting and might not be such an edge case for developers 😉. I won't have time to work on this until at least next week or the week after, so if you're able to take a whack at it, awesome! 

I'm curious what occurs when this happens/what output you see?

KennethNielsen:
@brandomr the case that I had was a day that had only one frame, that ended after midnight. In that case the aggregate for that day was empty.

KennethNielsen:
Hmm. As it turns out, this is a special case of ``report`` with the same start and end date not returning any output for en event that starts that day but doesn't end there. This in turn is due to ``Frames.span``, ``Frames.filter`` and ``Span.__contains__`` not doing that. This means that we are all the way down to the most basic objects and behaviour.

So at that point there are basically two routes:

1) Decide that for ``report``, this is intended behaviour... and implement ``aggregate`` without the ``report`` functionality. (This is a lot of work).
2) At some level (``Frames.filter`` and probably ``report``), introduce an (internal) option to concerning frames that cross the boundaries. If this options is active, it will then replace the frames in question with pseudo frames with timespans that ends at the boundary.

I tend to lean towards 2, especially considering that it can be implemented cleanly and without affecting any other functionality, but I would like some opinions before I start. What do you think @brandomr @jmaupetit?

jmaupetit:
I vote 2!

loonies:
I just started using 1.7 and this is annoying. At first I thought I'm doing something wrong.

brandomr:
@loonies thanks for bumping this! @jmaupetit I agree on option 2. I'll take a look into fixing this next week. I'm glad to see that people are finding the feature useful (minus this bug)

KennethNielsen:
@brandomr I hope you don't mind, but since I didn't see a PR I imagined you were busy, so I tried my hand at a fix. If you are already in progress with it, then if nothing else, it can serve as something to compare with.

[...]

PR #277, opened 2019-04-28
Add support for returning frames partially within span in filter and…

KennethNielsen:
… wire it up for aggregate

This is an attempt at a fix for #248. It adds an ``include_partial_frames`` argument to ``Frames.filter`` which if set True, and given a span, will return pseudo frames for frames that cross the span boundaries or for frames that envelope the entire span. The pseudo frames are copies of the original frames but where the start and/or endpoint has been edited to show **the part** of the frame that is within the span.

The ``include_partial_frames`` argument, is a keyword argument with default value False, which means that backwards compatibility should be kept for all other function until it can be decided if they need partial frames as well. For now, it is only wired up for aggregate.

Critique welcome:
 * I'm not quite happy with the name of the argument, suggestions welcome
 * This somewhat removed the elegance of the old implementation of the filter function and it is possible that the _partially in_ semantics can be done cleverer. Let me know.

I'm happy to try and do a test for it, but I want to know if you are happy with the form before I do that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improve Arrow 0.15.0 support after changes in `arrow.get()` behavior (#296)
 - Watson now suggests correct command if users make small typo (#318)
 
+
 ### Fixed
 
 - Stylize prompt to create new project or tag (#310).
 - Aggregate calculates wrong time if used with `--current` (#293)
 - The `start` command now correctly checks if project is empty (#322)
+- Aggregate ignores frames that crosses aggreagate boundary (#248)
 - The `report` and `aggregate` commands with `--json` option now correctly
   encode Arrow objects (#329)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improve Arrow 0.15.0 support after changes in `arrow.get()` behavior (#296)
 - Watson now suggests correct command if users make small typo (#318)
 
-
 ### Fixed
 
 - Stylize prompt to create new project or tag (#310).
--- a/watson/frames.py
+++ b/watson/frames.py
@@ -64,6 +64,9 @@ class Span(object):
         self.start = start.floor(self.timeframe)
         self.stop = stop.ceil(self.timeframe)
 
+    def overlaps(self, frame):
+        return frame.start <= self.stop and frame.stop >= self.start
+
     def __contains__(self, frame):
         return frame.start >= self.start and frame.stop <= self.stop
 
@@ -150,17 +153,33 @@ class Frames(object):
         ignore_projects=None,
         ignore_tags=None,
         span=None,
+        include_partial_frames=False,
     ):
-        return (
-            frame for frame in self._rows
-            if (projects is None or frame.project in projects) and
-               (ignore_projects is None
-                   or frame.project not in ignore_projects) and
-               (tags is None or any(tag in frame.tags for tag in tags)) and
-               (ignore_tags is None
-                   or all(tag not in frame.tags for tag in ignore_tags)) and
-               (span is None or frame in span)
-        )
+
+        for frame in self._rows:
+            if projects is not None and frame.project not in projects:
+                continue
+            if ignore_projects is not None and\
+               frame.project in ignore_projects:
+                continue
+
+            if tags is not None and not any(tag in frame.tags for tag in tags):
+                continue
+            if ignore_tags is not None and\
+               any(tag in frame.tags for tag in ignore_tags):
+                continue
+
+            if span is None:
+                yield frame
+            elif frame in span:
+                yield frame
+            elif include_partial_frames and span.overlaps(frame):
+                # If requested, return the part of the frame that is within the
+                # span, for frames that are *partially* within span or reaching
+                # over span
+                start = span.start if frame.start < span.start else frame.start
+                stop = span.stop if frame.stop > span.stop else frame.stop
+                yield frame._replace(start=start, stop=stop)
 
     def span(self, start, stop):
         return Span(start, stop)
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -818,6 +818,40 @@ def test_report_current(mocker, config_dir):
     assert len(report['projects']) == 0
 
 
+@pytest.mark.parametrize(
+    "date_as_unixtime,include_partial,sum_", (
+        (3600 * 24, False, 0.0),
+        (3600 * 48, False, 0.0),
+        (3600 * 24, True, 7200.0),
+        (3600 * 48, True, 3600.0),
+    )
+)
+def test_report_include_partial_frames(mock, watson, date_as_unixtime,
+                                       include_partial, sum_):
+    """Test report building with frames that cross report boundaries
+
+    1 event is added that has 2 hours in one day and 1 in the next. The
+    parametrization checks that the report for both days is empty with
+    `include_partial=False` and report the correct amount of hours with
+    `include_partial=False`
+
+    """
+    content = json.dumps([[
+        3600 * 46,
+        3600 * 49,
+        "programming",
+        "3e76c820909840f89cabaf106ab7d12a",
+        ["cli"],
+        1548797432
+    ]])
+    mock.patch('%s.open' % builtins, mock.mock_open(read_data=content))
+    date = arrow.get(date_as_unixtime)
+    report = watson.report(
+        from_=date, to=date, include_partial_frames=include_partial,
+    )
+    assert report["time"] == pytest.approx(sum_, abs=1e-3)
+
+
 # renaming project updates frame last_updated time
 def test_rename_project_with_time(watson):
     """
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -826,7 +826,7 @@ def test_report_current(mocker, config_dir):
         (3600 * 48, True, 3600.0),
     )
 )
-def test_report_include_partial_frames(mock, watson, date_as_unixtime,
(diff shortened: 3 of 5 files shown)


watson/frames.py
```diff
@@ -143,13 +143,28 @@ def new_frame(self, project, start, stop, tags=None, id=None,
...
-        return (
-            frame for frame in self._rows
-            if (projects is None or frame.project in projects) and
-               (tags is None or any(tag in frame.tags for tag in tags)) and
-               (span is None or frame in span)
-        )
+    def filter(self, projects=None, tags=None, span=None,
+               include_partial_frames=False):
+        for frame in self._rows:
+            if not (projects is None or frame.project in projects):
```
k4nar:
```suggestion
            if projects is not None and frame.project not in projects:
```

Would be easier to read IMO :) .

watson/frames.py
```diff
@@ -143,13 +143,28 @@ def new_frame(self, project, start, stop, tags=None, id=None,
...
+                continue
+            if not (tags is None or any(tag in frame.tags for tag in tags)):
```
k4nar: In the same spirit:

```suggestion
            if tags is not None and not any(tag in frame.tags for tag in tags)):
```

watson/frames.py
```diff
@@ -143,13 +143,28 @@ def new_frame(self, project, start, stop, tags=None, id=None,
...
+                continue
+            if span is None or frame in span:
+                yield frame
+                continue
```
k4nar:
```python
if span is not None and frame not in span:
    continue

yield frame
```

watson/frames.py
```diff
@@ -143,13 +143,28 @@ def new_frame(self, project, start, stop, tags=None, id=None,
...
+            # frames that are *partially* within span or reaching over span
+            if span is not None and include_partial_frames:
+                if frame.start < span.start and frame.stop > span.stop:
+                    yield frame._replace(start=span.start, stop=span.stop)
+                elif (frame.start < span.start and
+                      (span.start < frame.stop < span.stop)):
+                    yield frame._replace(start=span.start)
+                elif ((span.start < frame.start < span.stop) and
+                      frame.stop > span.stop):
+                    yield frame._replace(stop=span.stop)
```
k4nar: I think this would work as well (replacing my previous suggestion as well):
```python
if span is not None and if frame not in span:
    if include_partial_frames:
        if (span.start < frame.start < span.stop or span.start < frame.stop < span.stop):
            frame = frame._replace(start=max(frame.start, span.start), stop=min(frame.stop, span.stop))
    continue

yield frame
```

davidag: @k4nar I think that if you don't yield the result of `frame._replace()` you're going to lose it, because of the `continue` after it. Also, I'm not sure if the `if` inside the first condition is what you intended (`...and if frame not...`).

davidag:
> Actually I wonder if this couldn't be solved using the `timeframe` argument of the Span object:
> 
> https://github.com/TailorDev/Watson/blob/2d861a41e29859a685916a37240c950b2de67d0c/watson/frames.py#L62
> . But I haven't looked much into it though.

I couldn't come up with anything useful using `timeframe` :(

`Span` could have another method to check if there is an overlap:

```
class Span(object):
    def __init__(self, start, stop, timeframe='day'):
        self.timeframe = timeframe
        self.start = start.floor(self.timeframe)
        self.stop = stop.ceil(self.timeframe)

    def overlap(self, frame):
        return frame.start <= self.stop and frame.stop >= self.start

    def __contains__(self, frame):
        return frame.start >= self.start and frame.stop <= self.stop
```
I believe that with this new method, the proposed `filter()` code could be simplified.

davidag:
This looks good to me @KennethNielsen :relaxed: 

Let me add some _suggestions_ for the next steps:
1. Rebase your branch, resolve conflicts and (automatically) verify tests still pass.
2. Add new tests for the changes in this PR.
3. Add changelog entry.

KennethNielsen:
@davidag great. I have rebased the branch on top of current master, added a test and updated the changelog. 2 minor things though.

1) The rebase was non-trivial (due to the ignore project and tags changes also in filter), so it would be great if someone would glance this over once more before merging.
2) flake8 doesn't pass due to a line changed in a test in https://github.com/TailorDev/Watson/commit/581a107f217bdd65f9b6032aa8acd6ba2cce8433. It is trivial to fix, but I didn't know what the policy was on mixing stuff in a PR, so I didn't fix it and therefore flake8 reports error. Let me know and I can just include it if need be.

CHANGELOG.md
```diff
@@ -7,10 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added support for including partial frames in `Watson.report` and use it in
+  `cli.aggregate` (#277)
```
davidag: It looks like Watson's changelog _favors_ more high level comments, avoiding names of classes/methods, probably targeting Watson's users more than Watson's developers.

KennethNielsen:
Hi @davidag and thanks for your testing and feedback.

>     * If we extend the report to include both days, we get two entries as expected: OK, except that the sum is not the same by one second (3h 12s != 3h 11s).

Yes. I noticed the 1s issue. As I recall it originates from the way `report` uses arrow. Something about when `from` and `to` is the same date (which is what aggregate does) then it will ask the arrow date for `floor` and `ceil` and `ceil` is something like `****-**-** 23:59:59` and so it loses a second. It might and probably is fixable, but my feeling is that it would be too much spacial casing for such a small benefit, compared to the really clean implementation there is now.

>     * But if we get a `report` of the first day, _the result is empty_, which is what got me confused.
> 
> 
> ```
> watson report --from 2019-09-13 --to 2019-09-13
> Fri 13 September 2019 -> Fri 13 September 2019
> ```
> 
>     * If we extend the timespan to include both days, everything is fine.
> 
> 
> ```
> $ watson report --from 2019-09-13 --to 2019-09-14
> Fri 13 September 2019 -> Sat 14 September 2019
> bonus - 3h 00m 12s
> ```
> 
> I understand the `aggregate` command was the target of this PR, but it might make sense to make `report` and `aggregate` behave the same, even if the issue you reported was only related to the latter. Is there anything I am missing here? thinking

No. You are not missing something. _I think_ `include_partial_frames` should be used in report as well if you think about it as a _bugfix_, but I was a little wary about suggesting behavioral changes to core functionality like report and I was thinking that I would take discussion in a different PR. If you think it is natural that it should apply to report as well, then I will definitely add it.

> 
> I've reviewed all changes again, specially the `Frames.filter()` method, and it looks good to me +1 clap

Thanks!

CHANGELOG.md
KennethNielsen: I will change it the next time I get a few minutes for this.

davidag:
Hi @KennethNielsen!

> Yes. I noticed the 1s issue. As I recall it originates from the way `report` uses arrow. Something about when `from` and `to` is the same date (which is what aggregate does) then it will ask the arrow date for `floor` and `ceil` and `ceil` is something like `****-**-** 23:59:59` and so it loses a second. It might and probably is fixable, but my feeling is that it would be too much spacial casing for such a small benefit, compared to the really clean implementation there is now.

That seems very reasonable to me. I don't think it's worth the effort and added complexity.

> No. You are not missing something. _I think_ `include_partial_frames` should be used in report as well if you think about it as a _bugfix_, but I was a little wary about suggesting behavioral changes to core functionality like report and I was thinking that I would take discussion in a different PR. If you think it is natural that it should apply to report as well, then I will definitely add it.

I think it's more coherent if `report` and `aggregate` behave the same and return partial intervals when frames partially overlap the requested timespan.

@jmaupetit what do you think? Maybe this could be a complication for people relying on the current `report` command behavior.

KennethNielsen:
@davidag on the other hand, relying on frames being completely left out is kind of an exotic use case I think

KennethNielsen:
@davidag actually, the more I thought about it, the more it makes sense to enable this for report per default as well. So I went ahead and made the change and rebased once more (trivial this time) so it is ready for merge.

KennethNielsen:
Ok. So it appears that this branch now has conflicts again. I will wait to resolve the conflicts until there is a decision on the default behavior mentioned above.

KennethNielsen:
@davidag who would need to approve the default behavior change mentioned?

davidag:
@KennethNielsen: You might want to rebase first, because there are a couple of conflicts in your PR. After that, @k4nar is the person you're looking for :wink:

KennethNielsen:
@davidag thanks. I rebased this already twice (one of them non-trivial) and I don't really think it is necessary to make the decision on the default behavior. Understand that I'm not complaining about the process or anything, I would just like to know that there **is** a path to having it merge before I put more effort into it.

@k4nar let me know if you have the time to make a decision about the default behavior on what to do with frames that cross intervals in aggregate and then I can do the rebase in no time.

davidag:
@KennethNielsen: Sure, I understand your concerns. Let's hope @k4nar will come back to this PR and give you some feedback :crossed_fingers:

KennethNielsen:
@k4nar I resolved the merge conflicts and fixed the tests. I hope you will have the time to look at this at some point.

KennethNielsen:
@k4nar great. I added a code test but it could definitely use be real tested. I will try and recruit someone from the issue to test it.

@emarthinsen, @brandomr @loonies would any of you 3 mind testing this branch?

KennethNielsen:
Oh you actually meant code coverage. I can see that merging is blocked. Should I do another rebase?

k4nar:
I think it's only blocked because you and I don't have the rights to merge. ping @jmaupetit :)

jmaupetit:
@KennethNielsen rebase your branch, and I'll merge it! :pray:

KennethNielsen:
@jmaupetit done

jmaupetit:
Yay :tada: Thanks again @KennethNielsen! I apologize for my lack of reactivity...

KennethNielsen:
@jmaupetit no problem. Thanks for merging.

Issue #278, opened 2019-04-29
Make adding past activities easier


varac:
Right now, adding a past activity needs a lot of typing:

    watson add --from "2019-04-24 10:00:00" --to "2019-04-24 11:00:00" project1 +task1

even when adding an activity from today.

It would be nice if you don't need to provide the full date+time, and watson would gracefully accept i.e. for today:

    watson add --from "10:00:00" --to "11:00:00" project1 +task1
    watson add --from "10" --to "11" project1 +task1

or sth like this:

    watson add --from "yesterday 10:00" --to "yesterday 11:00" project1 +task1

jmaupetit:
I do agree with you, this would be lovely! Do you have time to work on a PR for this?

varac:
@jmaupetit Sorry, not the time nor the python skills needed :/

jmaupetit:
No worries, we'll find time/a volunteer to implement this feature.

davidag:
[Arrow's date/time creation alternatives](https://arrow.readthedocs.io/en/latest/#creation) don't seem to accept a ["humanized" string](https://arrow.readthedocs.io/en/latest/#humanize) as input.

I imagine that it's far easier to generate a localized and humanized date from a given date, than it is to parse one into a date. This might be a good start point to search on the Internet.

jdckr:
I have issued PR #282 which adds possibility to use times only or one datetime and a time. Check documentation inside PR for more information.

prat0088:
I just tried out watson and encountered this issue within the first minute.  It's so very cumbersome to add time due to the verbose date format requirements.  

I would love to help out getting this issue through.  It looks like there are a series of PRs in various states:

- #282 Adds time-only support and tests,, but it's stalled since May 22nd, 2019
- #316 Adds unit tests for the date format, but it hasn't seen activity since Oct 7th, 2019

@davidag What's needed to get these PRs merged and this issue closed?  Is there anything I can help out with?

davidag:
@prat0088 PR #316 should be ready to be merged. When that's merged, PR #328 should go next making date input a bit more usable.

corbolais:
Without digging into the PRs given above, can I use the following?

    watson add -f "todayT03:45"
    watson add -f "todayT03h45"
    watson add -f "yesterdayT..."
    watson add -f "todayTnoon"

etc.

davidag:
@corbolais No, that syntax is not permitted.

michaelbeaumont:
[...]

PR #282, opened 2019-05-12
Add time only

jdckr:
Improve usability of watson add (issue #278)

### DIFF
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -1013,13 +1013,30 @@ def frames(watson):
     for frame in watson.frames:
         click.echo(style('short_id', frame.id))
 
+def isTime(date_time_string):
+    match = re.search("^\d{1,2}:\d{1,2}$", date_time_string)
+    return match != None
+
+def isDateTime(date_time_string):
+    match = re.search("^\d{4}-\d{1,2}-\d{1,2}\s\d{1,2}:\d{1,2}$", date_time_string)
+    return match != None
+
+def getDateTimeToday(time_string):
+    date_today = arrow.now().floor("day")
+    return getMergedDateTime(date_today, time_string)
+
+def getMergedDateTime(date_time, time_string):
+    hours, minutes = time_string.split(":")
+    return date_time.shift(hours=int(hours), minutes=int(minutes))
 
 @cli.command(context_settings={'ignore_unknown_options': True})
 @click.argument('args', nargs=-1)
-@click.option('-f', '--from', 'from_', required=True, type=Date,
-              help="Date and time of start of tracked activity")
-@click.option('-t', '--to', required=True, type=Date,
-              help="Date and time of end of tracked activity")
+@click.option('-f', '--from', 'from_', required=True, 
+              help="Date and time of start of tracked activity. "
+              "If date is omitted todays date is used.")
+@click.option('-t', '--to', required=True,
+              help="Date and time of end of tracked activity. "
+              "If date is omitted todays date is used.")
 @click.option('-c', '--confirm-new-project', is_flag=True, default=False,
               help="Confirm addition of new project.")
 @click.option('-b', '--confirm-new-tag', is_flag=True, default=False,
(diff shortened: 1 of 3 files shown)


jmaupetit:
Looks like we finally have our first CLI tests :wink:

k4nar:
Do you know why Github is displaying such a long diff for ` docs/user-guide/commands.md`? Did you by any chance generate it on Windows? If so, I think there is a way to avoid committing the `\r\n`: https://help.github.com/en/articles/dealing-with-line-endings

watson/cli.py
```diff
@@ -1013,13 +1013,30 @@ def frames(watson):
...
+    match = re.search("^\d{4}-\d{1,2}-\d{1,2}\s\d{1,2}:\d{1,2}$", date_time_string)
+    return match != None
+
+def getDateTimeToday(time_string):
+    date_today = arrow.now().floor("day")
+    return getMergedDateTime(date_today, time_string)
+
+def getMergedDateTime(date_time, time_string):
+    hours, minutes = time_string.split(":")
+    return date_time.shift(hours=int(hours), minutes=int(minutes))
```
k4nar: Those functions should probably go to the `utils` module (and be unit tested ;) ).

watson/cli.py
```diff
@@ -1053,6 +1071,25 @@ def add(watson, args, from_, to, confirm_new_project, confirm_new_tag):
...
+    # convert to todays datetime
+    if isDateTime(from_) and isDateTime(to):
+        from_ = arrow.get(from_).replace(tzinfo='local')
+        to = arrow.get(to).replace(tzinfo='local')
+    elif isDateTime(from_) and isTime(to):
+        from_ = arrow.get(from_).replace(tzinfo='local')
+        to = getMergedDateTime(from_.floor("day"), to)        
+    elif isTime(from_) and isTime(to):
+        from_ = getDateTimeToday(from_)
+        to = getDateTimeToday(to)
```
k4nar: I wonder if Arrow doesn't provide something to help with that. It feels a bit tedious to introduce our own regexp to parse the date & times.

jdckr:
> 
> 
> Do you know why Github is displaying such a long diff for ` docs/user-guide/commands.md`? Did you by any chance generate it on Windows? If so, I think there is a way to avoid committing the `\r\n`: https://help.github.com/en/articles/dealing-with-line-endings

Yes I am on Windows. I use checkout-as-is and commit-as-is which suits most cases very well. Have pushed a new commit fixing the line endings.

watson/cli.py
jdckr: I have checked the api and didn't find s.th. appropriate. But I am open for suggestions.

watson/cli.py
jdckr: Good point. 5729d3e71b

watson/cli.py
davidag: I've been looking into it and there's something I don't understand: why are there two different click types `Date` and `Time` in `cli.py`?

It turns out that both accept the full date, but using different formats!

* `TimeParamType` (aka, `Time`) is only used in `watson stop`:
```
$ watson stop --at 09:00
$ watson stop --at 2019-05-23T12:00
```
* `DateParamType` (aka, `Date`) is used un every other place where a date is required:
```
$ watson report -f 2019-05-23
$ watson report -f "2019-05-23 12:00"
```

In my opinion, the following changes might solve issue #278 in a simpler way:
1. Rename `Date` to `DateTime` which is more appropriate to what it does. Of course, rename also the corresponding `ParamType` classes.
2. Use only `DateTime` to unify datetime input management for the user. This means `watson stop --at` date format changes, but I think it's OK.
3. Allow `DateTime` to receive time only dates, which is precisely what the `Time` type does. You could reuse the code.
4. Remove `Time` and `TimeParamType`.

Please, remember that I'm not a reviewer of this project, so I might be missing important stuff!! :-) what do you think @k4nar?

jdckr: I like the idea. I didn't dive deep so I was not aware of TimeParamType which is basically doing what I implemented to handle time only.
I also couldn't agree more that the api is not consistent for the different cmds and it felt like adding another balcony.
The only drawback I see so far is that it will be hard to handle the case where we want to reuse the date specified for --from for --to as the parameter processing would be done without the knowledge about the other parameter.

PR #283, opened 2019-05-13
Status: elapsed time in minutes, not humanize()

raphaelfournier:
I worked on a widget for the AwesomeWM window manager, which displays watson's status with an icon (if a frame is started, it's a "play" icon, if not it's a "stop" icon, like a music player), surrounded by a (partial) circle which is a (circular) progress bar (reaching the full circle after an hour).
I needed the value of the spent time in the frame, but parsing the result of "humanize()" (given by the `status` command) is not very convenient. I copied the current flag code and wrote a slightly different version, which outputs the duration with a single number, in minutes (which could then be tweaked in the widget to account for hours or days).

A better version would probably merge the `-e` and `-m` flag to allow a formatting string.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ pip-log.txt
 *.sublime-workspace
 *.swp
 *.sw[po]
+data
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -329,8 +329,10 @@ def cancel(watson):
               help="only show tags")
 @click.option('-e', '--elapsed', is_flag=True,
               help="only show time elapsed")
+@click.option('-m', '--elapsedminutes', is_flag=True,
+              help="only show time elapsed in minutes")
 @click.pass_obj
-def status(watson, project, tags, elapsed):
+def status(watson, project, tags, elapsed, elapsedminutes):
     """
     Display when the current project was started and the time spent since.
 
@@ -371,6 +373,12 @@ def status(watson, project, tags, elapsed):
     if elapsed:
         click.echo(u"{}".format(
             style('time', current['start'].humanize())
+        ))
+        return
+
+    if elapsedminutes:
+        click.echo(u"{}".format(
+            style('time', current['start'])
         ))
         return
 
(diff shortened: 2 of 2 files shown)


jmaupetit:
> A better version would probably merge the -e and -m flag to allow a formatting string.

Yep.

0x7b1:
Hey man, did you found some clean solution to this? I also have this inconvenience in my polybar module

raphaelfournier:
Well, the solution that I propose *is* clean, even if it adds a new flag. So, for now, I think it could be accepted _as is_. Or, if I have to find time to merge `-e` and `-m` flags, it will have to wait a few weeks, probably.

jmaupetit:
We'll wait for it then :wink:

jmkerr:
It would be nice to have flags `-h, -H, --human-readable, --no-human-readable` in line with the rest of the interface and other tools. This flag would generalise better to other situations where, for example uuids are shortened for human consumption. 

For maximum non-human-readability and other reasons, the time should be returned in seconds.

raphaelfournier:
> It would be nice to have flags `-h, -H, --human-readable, --no-human-readable` in line with the rest of the interface and other tools. This flag would generalise better to other situations where, for example uuids are shortened for human consumption.
You're right, it's a great idea. The `-h` flag may, however, conflict with `-h, --help`, we'll have to be careful.

> For maximum non-human-readability and other reasons, the time should be returned in seconds.

You're right again, it was very hackish to have the time returned in minutes, seconds is way better.

.gitignore
```diff
@@ -39,3 +39,4 @@ pip-log.txt
 *.sublime-workspace
 *.swp
 *.sw[po]
+data
```
jmaupetit: :thinking: this is too specific to your current environment, it should not be globally ignored IMO.

watson/cli.py
```diff
@@ -374,6 +376,12 @@ def status(watson, project, tags, elapsed):
         ))
         return
 
+    if elapsedminutes:
+        click.echo(u"{}".format(
+            style('time',str(int((arrow.utcnow() - current['start']).seconds/60)))
```
jmaupetit:
```suggestion
            style('time', str(int((arrow.utcnow() - current['start']).seconds / 60)))
```

We should use Black to format / check the whole project code.

Issue #293, opened 2019-07-04
aggregate calculates wrong time if used with `--current`


nesmyslny:
If `aggregate` is used with `--current` and it takes more than the current day into account, the calculation of the current aggregated day is wrong.  
Example:
```sh
$ watson status
Project xy started 2 hours ago (2019.07.04 10:26:45+0200)


$ watson log -dc
Thursday 04 July 2019 (2h 17m 31s)
        current  10:26 to 12:44   2h 17m 31s  xy


$ watson report -dc
Thu 04 July 2019 -> Thu 04 July 2019

xy - 2h 17m 50s


$ watson aggregate -c --from 2019-07-03
Wed 03 July 2019 - 7h 17m 42s
  xy - 7h 17m 42s

Thu 04 July 2019 - 4h 36m 51s
  xy - 4h 36m 51s
```
Expected result of *Thu 04 July 2019*: 2h 17m ??s  
Expected result of *Thu 04 July 2019*: 4h 36m 51s

jmkerr:
The current frame is added once for [every day the report is generated](https://github.com/TailorDev/Watson/blob/master/watson/watson.py#L464).

nesmyslny:
Thanks for looking at it!

PR #294, opened 2019-07-08
Fixes multiple inclusions of the current frame in aggregate commands …

jmkerr:
Possible solution for #293.

### DIFF
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -795,6 +795,33 @@ def test_report(watson):
         watson.report(arrow.now(), arrow.now(), tags=["A"], ignore_tags=["A"])
 
 
+def test_report_current(config_dir):
+    watson = Watson(
+        current={'project': 'foo', 'start': arrow.now().shift(hours=-1)},
+        config_dir=config_dir
+    )
+
+    _ = watson.report(
+        arrow.now(), arrow.now(), current=True, projects=['foo']
+    )
+    report = watson.report(
+        arrow.now(), arrow.now(), current=True, projects=['foo']
+    )
+    assert len(report['projects']) == 1
+    assert report['projects'][0]['name'] == 'foo'
+    assert report['projects'][0]['time'] == pytest.approx(3600, rel=1e-2)
+
+    report = watson.report(
+        arrow.now(), arrow.now(), current=False, projects=['foo']
+    )
+    assert len(report['projects']) == 0
+
+    report = watson.report(
+        arrow.now(), arrow.now(), projects=['foo']
+    )
+    assert len(report['projects']) == 0
+
+
 # renaming project updates frame last_updated time
 def test_rename_project_with_time(mock, watson):
     """
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Stylize prompt to create new project or tag (#310).
+- Aggregate calculates wrong time if used with `--current` (#293)
 
 ## [1.8.0] - 2019-08-26
 
--- a/watson/watson.py
+++ b/watson/watson.py
@@ -467,13 +467,13 @@ class Watson(object):
         if from_ > to:
             raise WatsonError("'from' must be anterior to 'to'")
 
-        if self.current:
-            if current or (current is None and
-                           self.config.getboolean(
-                               'options', 'report_current')):
-                cur = self.current
-                self.frames.add(cur['project'], cur['start'], arrow.utcnow(),
-                                cur['tags'], id="current")
+        if current is None:
+            current = self.config.getboolean('options', 'report_current')
+
+        if self.current and current:
+            cur = self.current
+            self.frames.add(cur['project'], cur['start'], arrow.utcnow(),
+                            cur['tags'], id="current")
 
         span = self.frames.span(from_, to)
 
@@ -486,6 +486,9 @@ class Watson(object):
             ),
             operator.attrgetter('project')
         )
+
+        if self.current and current:
+            del self.frames['current']
 
         total = datetime.timedelta()
 


jmaupetit:
Thank you @jmkerr ! Can you provide tests for this fix please?

jmkerr:
Thank you for Watson, I'm happy to help!

I'm not exactly sure where to best handle the deduplication of the `current` frame. Maybe it should be part of the behaviour of the `frames` data structure. Is it ever really useful to ignore the current frame?

For now, this is a simple and fast fix.

jmaupetit:
It makes sense indeed.

jmkerr:
> Maybe instead we should remove the "current" frame at the end of the `report` function? It looks like a side effect.

You're right. I have changed it to always initialize `current` to keep track of the extra frame and remove it as soon as possible.

tests/test_watson.py
```diff
@@ -794,6 +794,12 @@ def test_report(watson):
     with pytest.raises(WatsonError):
         watson.report(arrow.now(), arrow.now(), tags=["A"], ignore_tags=["A"])
 
+    watson.start('bar')
+    _ = watson.report(arrow.now(), arrow.now(), current=True)
+    watson.stop()
+
+    assert watson.frames['id'].count('current') == 0
```
davidag: It looks like you're testing the implementation here (i.e. the current frame has been deleted after report() finishes), but maybe it would be even more useful to test the actual report data returned?

The question I'd like to get answered by the test is this: _are we sure that the current frame is not being deleted too soon in the report() method?_

The `watson.py` implementation looks fine to me :+1:

davidag: @jmkerr I'm not sure if I made myself clear. If you need further clarification just let me know.

@k4nar Sorry for stepping in here, I thought it might be useful to move the PR forward... what do you think about my comment?

jmkerr:
Thank you @davidag. Since this commit changes the implementation, it should include the tests you mentioned. The tests now assert whether the current frame is included in the report
- when it should be 
- when it should not be
- by default

tests/test_watson.py
```diff
@@ -794,6 +794,26 @@ def test_report(watson):
...
+    report = watson.report(
+        arrow.now(), arrow.now(), current=True, projects=['foobaz']
+    )
+    assert len(report['projects']) == 1
+    assert watson.frames['id'].count('current') == 0
+
+    report = watson.report(
+        arrow.now(), arrow.now(), current=False, projects=['foobaz']
+    )
+    assert len(report['projects']) == 0
```
davidag: You might want to test that the `foobaz` project is included in the report, not only that one project is included (even if we know that it has to be it :relaxed:).

tests/test_watson.py
davidag: I'll try to explain my point further: I think you should solely rely on `Watson` class methods to test your assumptions on the behavior of the `Watson` class.

As I see it, by accessing the `Frames` object and asserting on its properties, you're effectively testing an implementation detail of the `Watson` class and adding complexity to the test itself.

What I would have done here is calling two or more times `watson.report()` and verify that you're always getting the expected report results.

What do you think? :relaxed:

tests/test_watson.py
```diff
@@ -794,6 +794,26 @@ def test_report(watson):
...
+
+    report = watson.report(
+        arrow.now(), arrow.now(), projects=['foobaz']
+    )
+    assert len(report['projects']) == 0
+
+    watson.stop()
```
davidag: Finally, something I missed in my initial review: this seems a great opportunity to split `test_report()` in multiple tests with more meaninful names. For example, your new tests could be inside a method named `test_report_current()`.

jmkerr:
Thanks again @davidag. You're right about not explicitly testing for side effects, but only the methods return value. The only way to detect the error in a report is by verifying the reported time.

I've updated the test (in a new function) to:
 - Construct `watson` with a current frame with a backdated start date.
 - Verify the reported time (up to 1%).

tests/test_watson.py
```diff
@@ -795,6 +795,33 @@ def test_report(watson):
...
+
+    _ = watson.report(
+        arrow.now(), arrow.now(), current=True, projects=['foo']
+    )
+    report = watson.report(
+        arrow.now(), arrow.now(), current=True, projects=['foo']
+    )
+    assert len(report['projects']) == 1
+    assert report['projects'][0]['name'] == 'foo'
+    assert report['projects'][0]['time'] == pytest.approx(3600, rel=1e-2)
```
davidag: I don't think you need to use `pytest.approx()` as there are no floating-point numbers involved. In fact, asserting that time equals `3600` is equivalent to what you did, as we are working with a granularity of seconds (`int`s). This would lead to a flaky test: if there was a change in seconds between building the `Watson` object and calling the second `watson.report()`, the assertion would fail.

A possible solution could be to assert that time is _less than_ `3600 * 2`, which is enough to test that the time is not duplicated. Any other ideas are more than welcomed! :relaxed: 

Finally, I would use one minute (60) instead of one hour (3600) in order to reduce the complexity of the test (unless there is a reason to use hours that I didn't notice).

It seems we're close to ending this PR :tada:

jmkerr: `report['projects'][0]['time']` is (like `timedelta.total_seconds()`) floating-point-valued with microsecond precision, and it usually takes a few milliseconds until the second report is generated.

The flakyness should be just as negligible at the current 36 seconds as it is at 3600 seconds tolerance in an automated test, but could be reduced by mocking `arrow.now` and `arrow.utcnow`.

davidag: > report['projects'][0]['time'] is (like timedelta.total_seconds()) floating-point-valued with microsecond precision, and it usually takes a few milliseconds until the second report is generated.

I didn't realize that, sorry!

I supposed they were seconds, because that's what is printed in the terminal. Nevertheless, it's the `utils.format_timedelta()` function that removes millisecond information: `seconds = int(delta.total_seconds())`. :+1: 

> The flakyness should be just as negligible at the current 36 seconds as it is at 3600 seconds tolerance in an automated test

You're completely right, there's no flakiness :relaxed:

jmkerr:
I've rebased and added a line to `CHANGELOG.md`.

PR #305, opened 2019-08-30
Remove `python2.7` from supported versions of Python

MaxG87:
Version 2.7 of Python soon will reach the end of its live. It already
hinders the application of new language features which are incompatible
with 2.7. Thus, stopping to support `python2.7.` is necessary.

All major distributions ship some version of the Python3 series already.
Thus, stopping to support `python2.7` should be relatively light in
trouble for users.

### DIFF
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -199,7 +199,7 @@ toto
 
 
 def test_empty_config(mocker, watson):
-    mocker.patch.object(ConfigParser, 'read', mock_read(u''))
+    mocker.patch.object(ConfigParser, 'read', mock_read(''))
     assert len(watson.config.sections()) == 0
 
 
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -10,15 +10,11 @@ import requests
 
 from watson import Watson, WatsonError
 from watson.watson import ConfigParser, ConfigurationError
-from watson.utils import PY2
 
 from . import mock_read, TEST_FIXTURE_DIR
 
 
-if not PY2:
-    builtins = 'builtins'
-else:
-    builtins = '__builtin__'
+builtins = 'builtins'
 
 
 @pytest.fixture
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -190,7 +190,7 @@ def test_empty_config_dir():
 
 
 def test_wrong_config(mocker, watson):
-    content = u"""
+    content = """
 toto
     """
     mocker.patch.object(ConfigParser, 'read', mock_read(content))
@@ -237,7 +237,7 @@ def test_start_two_projects(watson):
 
 
 def test_start_default_tags(mocker, watson):
-    content = u"""
+    content = """
 [default_tags]
 my project = A B
     """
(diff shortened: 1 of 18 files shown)


jmaupetit:
That's a breaking change that should stand in Watson 2.0.0 roadmap!

I have no idea how many Watson's users are still using python 2...

MaxG87:
I have to admit that I created this PR mainly because I was so annoyed of having to import stuff in a way Python2.7 can digest. Anyhow, Python3.5 is available even in Debian Stretch, Pyhon3.6 in RedHat 7.

k4nar:
https://pythonclock.org :smile:

This would be a great thing, but as @jmaupetit mentioned it should probably be released in a 2.0.

tests/test_watson.py
```diff
@@ -21,10 +20,7 @@
         )
     ) / 'resources'
 
-if not PY2:
-    builtins = 'builtins'
-else:
-    builtins = '__builtin__'
+builtins = 'builtins'
```
k4nar: I think we don't need this var anymore, we can use `'builtins'` directly where we need.

MaxG87: Done.

KennethNielsen:
I just wanted to add, that whether by accident or purpose, that last time I ran the test suite, there were also python 2 problems.

MaxG87:
On 19-10-22 08:57:17, David Alfonso wrote:
>davidag requested changes on this pull request.
>
>Some other things that apply to Watson:
>- Encoding declarations at the beginning of files can be removed.
>- All ImportError try/except blocks can be removed.
>- Add `python_requires` in setup.py.
>- The `u` prefix in string literals can be removed as there are only Unicode literals in Python3.

I agree with this. I will extend the PR soon.

MaxG87:
Now with 1.9.0 out (thank you very much), maybe its time to apply this patch? I suspect plenty of other benefits to be reachable once we get this, e.g. typing.NamedTuple or dataclasses.

Maybe one could directly move to 3.6?

k4nar (dismissed):

MaxG87:
Are there any updates here? I would like to contribute a bit more but I am annoyed by having to pay attention to Python 2.7 (and, to lesser extend, to Python 3.5). Even Python 3.5 has reached EOL.

k4nar (dismissed):

watson/cli.py
```diff
@@ -136,7 +134,7 @@ def cli(ctx, color):
 
     You just have to tell Watson when you start working on your
     project with the `start` command, and you can stop the timer
-    when you're done with the `stop` command.
+    when yo're done with the `stop` command.
```
jmaupetit, line 137: oops, looks like `sed` found a false-positive :wink: 

```suggestion
    when you're done with the `stop` command.
```

watson/version.py
```diff
@@ -1,3 +1 @@
-# -*- coding: utf-8 -*-
-
-version = "1.10.0"
+version = "1.10.0"
```
jmaupetit, line 1:
```suggestion
version = "1.10.0"

```

jmaupetit (commented):
I'll merge this once rebased (and fixed). It's a major step. Thanks! :pray:

jmaupetit:
Thanks @MaxG87 :pray: I forgot to ask you one last thing: can you add this major change in the CHANGELOG please?

PR #307, opened 2019-08-30
Set WATSON_DIR on virtualenv activation in fish script

nesmyslny:
Hey!

Just a small change to make dev experience with fish shell a bit more comfortable. Not sure if you like it or want to support fish (or have fish specifics in the docs). Would be nice to have this merged, but no strong feelings if you close the request :smile: 

This commit includes:
* Adds `set -x WATSON_DIR [...]` to `.venv/bin/activate.fish`, so that its also set when using fish specific activation script (similar to posix activation script).
* Adds hint for fish users in the docs about `activate.fish`.

If you prefer, I can split it into two commits, of course.


nesmyslny:
Thanks @jmaupetit!

PR #308, opened 2019-09-02
Add decimal formatting for durations

64r:
Hi,

I've added functionality that allows timedeltas to be formatted as a decimal number instead of hours, minutes and seconds.
This is useful for users that want to calculate a price based on an hourly rate. For example `watson aggreate` now shows `4.50h` if you've tracked 4 hours and 30 minutes.

This feature is configurable via `watson config`. You can also set the decimal precision used for formatting the numbers.

I've also added documentation and tests for this feature.

### DIFF
--- a/watson/utils.py
+++ b/watson/utils.py
@@ -8,6 +8,7 @@ import os
 import shutil
 import sys
 import tempfile
+
 try:
     from StringIO import StringIO
 except ImportError:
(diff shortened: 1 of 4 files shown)


watson/utils.py
```diff
@@ -84,11 +85,15 @@ def _style_short_id(id):
         return fmt(element)
 
 
-def format_timedelta(delta):
+def format_timedelta(watson, delta):
```
jmaupetit: This is breaking the API, what about the following instead?

```suggestion
def format_timedelta(delta, as_decimal=False):
```

64r: That's true. I can change it to the way you suggested. Would you prefer if I made this a file-local variable in cli.py or do the call in every method that uses format_timedelta?

jmaupetit: I would prefer not having a separated method to format timedelta but a single entrypoint that does the heavy lifting. This should be sufficient for such a small piece of code and its testability. WDYT @k4nar?

PR #318, opened 2019-09-20
Enable git-like did-you-mean feature

hiiwave:
As watson is a command-intensive application, I think it would be a plus if it comes with did-you-mean feature powered by [click-didyoumean](https://github.com/click-contrib/click-didyoumean), e.g.
```
$ watson strat
Usage: watson [OPTIONS] COMMAND [ARGS]...
Try "watson --help" for help.
Error: No such command "strat".
Did you mean one of these?
start
status
restart
```

What do you think about this?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 arrow!=0.11,!=0.12.0
 click
 requests
+click-didyoumean
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 arrow!=0.11,!=0.12.0
 click
-requests
 click-didyoumean
+requests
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,3 +1,4 @@
 arrow!=0.11,!=0.12.0
 click
 requests
+click-didyoumean
```
davidag: I would put both click dependencies togeter.

hiiwave: I'm neutral to the style here and let's see if maintainers have any preference.

jmaupetit: I like when things are alphabetically sorted :wink:

jmaupetit:
I :heart: it!

PR #323, opened 2019-09-23
Allow to use fuzzed data in tests

davidag:
**List of changes**
- Change fuzzer.py to also support receving a path as argument where fuzzed data will be generated. Without arguments, the behavior doesn't change.
- New pytest fixture watson_df to support a non-empty config directory using pytest-datafiles. This allows to work on a copy of the original data files on each test.
- Create an empty directory where fuzzed data will be automatically generated in CI, as specified in tox.ini.
- Make TEST_FIXTURE_DIR global in the tests module, as it will be required by any test that wants to mark paths/files for pytest-datafiles.

**How to use fuzzed data in a test through the new fixture**

    @pytest.mark.datafiles(TEST_FIXTURE_DIR / 'fuzzer_data')
    def test_something(watson_df):
        pass

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,10 @@ deps =
     mock
     pytest-datafiles
     pytest-mock
-commands = py.test -vs tests/
+commands =
+    python scripts/fuzzer.py tests/resources/fuzzer_data/
+    py.test -vs tests/
+
 usedevelop = True
 
 [testenv:flake8]
--- a/scripts/fuzzer.py
+++ b/scripts/fuzzer.py
@@ -8,41 +8,52 @@ import sys
 
 from watson import Watson
 
-if not os.environ.get('WATSON_DIR'):
-    sys.exit(
-        "This script will corrupt Watson's data, please set the WATSON_DIR "
-        "environment variable to safely use it for development purpose."
-    )
-
-watson = Watson(config_dir=os.environ.get('WATSON_DIR'),
-                frames=None,
-                current=None)
-
-projects = [
+FUZZER_PROJECTS = [
     ("apollo11", ["reactor", "module", "wheels", "steering", "brakes"]),
     ("hubble", ["lens", "camera", "transmission"]),
     ("voyager1", ["probe", "generators", "sensors", "antenna"]),
     ("voyager2", ["probe", "generators", "sensors", "antenna"]),
 ]
 
-now = arrow.now()
 
-for date in arrow.Arrow.range('day', now.shift(months=-1), now):
-    if date.weekday() in (5, 6):
-        # Weekend \o/
-        continue
-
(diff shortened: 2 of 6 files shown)


tox.ini
```diff
@@ -9,7 +9,10 @@ deps =
     mock
     pytest-datafiles
     pytest-mock
-commands = py.test -vs tests/
+commands =
+    python scripts/fuzzer.py tests/resources/fuzzer_data/
```
k4nar: Does that mean that the tests will fail if someone uses `pytest` directly and doesn't launch this command first?

davidag: Probably the tests would fail in that case. As there are no tests using fuzzed data now, we can't say 100%, but I see where you're going with this :wink: 

Possible solutions:
1. Suppose fuzzed data is fixed in some point in time and just upload it to the git repository. The need to recreate the data would vanish, though tests would have less potential.
2. Modify [Run the tests](https://tailordev.github.io/Watson/contributing/hack/#run-the-tests) doc, specifying some new steps to mitigate errors running the tests.
3. Create another target in the `Makefile` for running the tests that includes the fuzzed data generation. Document that in the hack doc :crossed_fingers: 
4. Check for fuzzed data existence inside fixture `watson_df`, before creating the `Watson` object, and make tests fail printing a proper error message.
5. Auto-generate the data from inside the tests by running an external process (`python scripts/fuzzer.py tests/resources/fuzzer_data`).
6. Auto-generate the data from inside the tests by directly running Python code. A new module could be created that would allow to generate fuzzed data. This would be used from `scripts/fuzzer.py` and from the tests. Tests which required different kinds of fuzzed data could be easily created by expanding this module.
7. _Put your option here_ :relaxed: 

What do you think?

davidag: @k4nar I couldn't come up with any other option. If you pointed me to what you think is best for the project, I'd be glad to modify the PR (or even close it if you don't see value in it! No hard feelings! :relaxed:)

k4nar: Option 6 looks the best to me. But I'm still not sure in which context we would like to use fuzzed data in the tests :) .

davidag: I was probably thinking in a way to have more complex data than what is currently used in the tests and `fuzzer.py` seemed like a good place to start.

Besides that, [fuzzed data is quite useful when testing](https://www.softwaretestingclass.com/fuzz-testing/) and I imagine that's the reason you created the aforementioned script in the first place.

But I understand your point and maybe it's better to leave this PR for whenever it makes sense to do some fuzzed testing :relaxed:

davidag:
Closing this PR for whenever it is appropriate to add some fuzz tests.

tox.ini
k4nar: Thank you for your time, we'll definitively reconsider this if there is a need in the future!

PR #330, opened 2019-10-16
Use Arrows for_json for JSON output

### LINKED ISSUE #329: The `report` and `aggregate` commands with `--json` option don't work any more
`watson report --json` and `watson aggregate --json` to fail with this error message:

> TypeError: Object of type Arrow is not JSON serializable

The regression was introduced in PR #313.

rbialon:
Arrow provides a `for_json()` function encoding Arrow objects for JSON output.
Using the default-parameter of `json.dumps()`, this function is invoked on Arrow objects which can then be encoded for JSON output.

This PR fixes #329


rbialon:
Sure, I added a short description to the changelog and provided some basic tests.

Let me know if the tests should cover more cases, I provided ones for the two that came to my mind.

Issue #309, opened 2019-09-05
watson edit: add --from, --to switches


abarger-bss:
One of the most common things I do in Watson is edit my current frame because I started or stopped watson well after I started or stopped my work. It would be nice if I could edit the dates for a frame directly in the CLI with --from or --to switches on the edit command.

jmaupetit:
Thanks for the report @abarger-bss! Do you have time to implement such feature and propose a PR? :pray:

benallan:
Hi, I actually was hoping to use these same options too so I opened a PR for them: #333. I hope you don't mind, @abarger-bss.

abarger-bss:
Not at all, I'm relieved. Sorry for ghosting on this issue - Python's not my daily tool so I would have had a hard time with this. Glad to hear the suggestion makes sense for others though.

I will add that since opening this issue I've had good results using `watson stop --at <time>` and `watson start --no-gap` to set frame times without a text editor.

benallan:
Thanks a lot for reminding me to check the `stop --at` option, I really needed to see which time formats it accepts. I also like using `stop --at`, but I often seem to be so forgetful that I need to edit the times for 2 or 3 frames back... So I suppose that new start/stop options for the edit command might still be useful.

PR #333, opened 2019-10-27
Add start and stop options to edit command

benallan:
As mentioned in #309, it's sometimes convenient to edit the start or stop times of a frame without opening an editor, so this PR adds two parameters to the `watson edit` command:
- `--start` sets the start time for a frame
- `--stop` sets the stop time for a frame

I chose the names "start" and "stop" rather than "from" and "to" since it matches the json shown in in the editor (and the names watson uses with frames internally).

Example usage:
```
watson edit --start "2019-10-27T14:30:00" --stop "2019-10-27T15:00:00"
watson edit --start "14:30" --stop "15:00"
watson edit --start "14:30" 
watson edit -2 --stop "15:00"
```

I rebuild the docs since the docstring and parameters for `edit` had changed, and it seems like a few other pending changes to the docs were also included (looks like mainly punctuation), so sorry if that causes any confusion. 

### TODO

**Update**: I've updated the PR to use TimeParamType instead which is a bit easier to use, so the comments below are obsolete. I found the `stop --at` command which already uses this syntax so it's also more consistent to use the same here. 

I'd quite like to be able to specify just a time on the command line, since typing the date is a bit of a hassle and is usually not needed since we could use the current date or keep the date the frame has unchanged. However for now I've used the DateParamType, which means we have to specify date and time. Using TimeParamType instead would mostly do what I was hoping as the user could choose either a time only or a time with date. But I'm not sure if the format is the most convenient: TimeParamType won't recognize spaces between the date and time ("2019-10-27 15:00:00"); it needs the letter "T" between them ("2019-10-27T15:00:00"). I had a quick search to see if TimeParamType was used elsewhere but couldn't find any other usages, so I'm not really sure if there's any existing precedent for this. I'd be grateful if anyone has any advice!

### DIFF
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -12,7 +12,7 @@
 Usage:  watson add [OPTIONS] [ARGS]...
 ```
 
-Add time for project with tag(s) that was not tracked live.
+Add time to a project with tag(s) that was not tracked live.
 
 Example:
 
@@ -53,7 +53,7 @@ If you are outputting to the terminal, you can selectively enable a pager
 through the `--pager` option.
 
 You can change the output format from *plain text* to *JSON* using the
-`--json` option or to *CSV* using the `--csv` option. Only one  of these
+`--json` option or to *CSV* using the `--csv` option. Only one of these
 two options can be used at once.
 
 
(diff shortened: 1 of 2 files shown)


benallan:
Closing this, because I later switched to use a workflow more like #388.

PR #326, opened 2019-09-30
Solve conflicts for #140

gkmngrgn:
image

Hi, I fixed the conflicts keeping @SpotlightKid 's changes and tested his changes. It seems working well. Do we need to do something else?

My config file was:
```
[style:project]
fg = black
bg = white

[style:tag]
fg = yellow
bg = white
```

Thanks.

PR #335, opened 2019-10-30
Style Customization Support #140 - #326

gkmngrgn:
@k4nar Unfortunately I can't reopen the old pull-request, it's a fixed version of #326.

### DIFF
--- a/watson/utils.py
+++ b/watson/utils.py
@@ -32,6 +32,7 @@ try:
 except NameError:
     text_type = str
 
+
 DEFAULT_STYLES = {
     'project': {'fg': 'magenta'},
     'tag': {'fg': 'blue'},
@@ -44,6 +45,16 @@ DEFAULT_STYLES = {
 
 def create_watson():
     return _watson.Watson(config_dir=os.environ.get('WATSON_DIR'))
+
+
+def get_styles_from_config():
+    ctx = click.get_current_context(silent=True)
+    if ctx is None:
+        return
+
+    config = ctx.obj.config
+    for name, style in DEFAULT_STYLES.items():
+        DEFAULT_STYLES[name] = config.getitems('style:%s' % name, style)
 
 
 def confirm_project(project, watson_projects):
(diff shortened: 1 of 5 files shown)


watson/utils.py
```diff
@@ -85,6 +85,11 @@ def _style_short_id(id):
     fmt = formats.get(name, {})
 
     if isinstance(fmt, dict):
+        try:
+            config = click.get_current_context().obj.config
+            fmt.update(config.getitems('style:%s' % name))
+        except (AttributeError, RuntimeError):
+            pass
```
k4nar: I wonder if that could be done somewhere during the startup (maybe in a lazy way?) instead of when styling every element. For long outputs with lots of formatting this might slow things down.

gkmngrgn: @k4nar hi, I added a function to load styles just on the startup:
https://github.com/TailorDev/Watson/pull/335/files#diff-53886fd7d4ba12dfad3dbb6c9716ca57R45-R52

Can you please review it again?

watson/utils.py
```diff
@@ -71,16 +91,11 @@ def _style_tags(tags):
...
+    formats.update({
         'tags': _style_tags,
-        'tag': {'fg': 'blue'},
-        'time': {'fg': 'green'},
-        'error': {'fg': 'red'},
-        'date': {'fg': 'cyan'},
         'short_id': _style_short_id,
-        'id': {'fg': 'white'}
-    }
+    })
```
k4nar: I think we should extract the nested functions & add them in `get_styles_from_config` in order to avoid a copy every time we call `style` :) .

watson/utils.py
```diff
@@ -32,6 +32,26 @@
...
+    'error': {'fg': 'red'},
+    'date': {'fg': 'cyan'},
+    'id': {'fg': 'white'}
+}
+
+
+def get_styles_from_config():
+    ctx = click.get_current_context(silent=True)
+    if ctx is None:
+        return
```
k4nar: Maybe we should take `ctx` in parameter instead?

watson/utils.py
gkmngrgn: @k4nar I think I need your review again. I solved the performance problem that you mentioned but I followed a slightly different way.

gkmngrgn:
@k4nar do I need to do something else in this ticket?

k4nar:
I guess it would be worth a CHANGELOG entry :) .

Otherwise I think we're good, but I don't have the rights to merge this, so we'll have to wait for @jmaupetit .

jmaupetit:
Yes, add a new entry to the CHANGELOG and I'll merge this @gkmngrgn 

:pray: @k4nar for the review

PR #252, opened 2019-02-09
WIP: Add log messages to frames 

jnsebgosselin:
In my opinion, there is no need for PR #115 to be blocking the implementation of this feature.

So as a first step, I've taken the work done in PR #119 and made it independent of PR #115. The work done in PR #115 can be completed at a later time.

PR #337, opened 2019-11-16
feat: add messages to time logs

### LINKED ISSUE #93: Add git commit-like comment on watson stop

prat0088:
Copied from @jnsebgosselin 's work on  #252 and rebased.

To those coming from the feature request issues, you can install this now to play around with by running by running

> pip install https://github.com/prat0088/Watson/archive/logmessage.zip

It has been working fine for me on Windows.  But please be warned this change hasn't yet been merged into master so there's no guarantees it will continue to work or that it won't corrupt your existing data.

Closes #93


prat0088:
@k4nar messages aren't yet displayed when running the log command.  I'm unsure the best way to display them.  If you have a preference please leave a comment over in #93 .

k4nar:
Maybe we could do this in another PR though? I know this feature doesn't make much sense without it, but it's a good first step.

prat0088:
@k4nar I agree, but @jmaupetit prefers to have a single, feature-complete change ([commented here](https://github.com/TailorDev/Watson/pull/252#issuecomment-553304645)).

prat0088:
At an rate, removing the WIP.

k4nar:
Oh sorry I totally forgot about this conversation which happened only 6 days ago :joy: . Let's wait for a consensus on the behavior of `watson log` then.

prat0088:
Almost there.  start, stop, and edit now work with messages.  You can pass a message to both start and stop.  

Remaining: display messages when running the log command.  They display in json and csv formats, but not plain.

prat0088:
@k4nar @jnsebgosselin I could use a second opinion:

I had been using -m/--message on start/stop, mirroring the git parameter of the same name.

I added options --messages --no-messages to log, aggregate, and report to control whether or not messages get displayed.  They really add a lot of noise, but are also useful in some cases, so I want to make sure users have can show and hide them when needed.

The problem is that aggregate and report already define -m to be --month, so I could define --message, but not -m.  I want to keep the message flag consistent across commands to make it more intuitive to use.  However -l (L) is already taken (--log-message), -s is already taken by --csv (--string, --summary), -t is already taken by --to (--text)....

I just thought of --note.  No commands are using -n yet.

**TL;DR**

What do you think about renaming -m/--message to -n/--note so that we can use -n on start, stop, log, aggregate, and report?  No, it's not the same as git's option, but I think a consistent abbreviation will make watson much easier to use, and the usage of -n is slightly different than git's -m anyway.

prat0088:
Update: Messages now get aggregated and displayed in report and aggregate, hidden by default.

prat0088:
@jnsebgosselin I'm going to go ahead with renaming -m and --message to -n and --note

prat0088:
note to self: there is a bug in watson aggregate --note.  It is expecting a parameter.  It doesn't function as a flag.

prat0088:
@k4nar This is ready for review.

prat0088:
@k4nar @jmaupetit It has been two months since I have heard from either of you on this request.  It is ready to merge.  What needs to happen next?

k4nar:
@prat0088 : Sorry, I'll try to do a final review soon :pray: .

watson/cli.py
```diff
@@ -227,6 +236,7 @@ def start(ctx, watson, confirm_new_project, confirm_new_tag, args, gap_=True):
 
     if project and watson.is_started and not gap_:
         current = watson.current
+        # TODO: log in error note
```
k4nar: I think we can drop the TODO, don't you think?

watson/cli.py
```diff
@@ -485,11 +516,13 @@ def status(watson, project, tags, elapsed):
               help="Format output in plain text (default)")
 @click.option('-g/-G', '--pager/--no-pager', 'pager', default=None,
               help="(Don't) view output through a pager.")
+@click.option('-n', '--note', 'show_notes', default=False, is_flag=True,
```
k4nar: I think it should be `--notes`

watson/watson.py
```diff
@@ -288,9 +291,14 @@ def stop(self, stop_at=None):
         if stop_at > arrow.now():
             raise WatsonError('Task cannot end in the future.')
 
+        if note is None:
+            note = old.get('note')
```
k4nar: So that means that if a not is passed with `watson stop`, it will override a previous note passed via `watson start`? Shouldn't we try to concatenate them somehow, or at least not silently drop it.

prat0088: I’m not sure what people would prefer: concatenating or dropping. Either way silently dropping is bad I agree.

How about printing a message that the old note from start is being overwritten, print it, the print the new one?  The user can edit the frame if they want to amend the note. It will be obvious and they will have a copy of the old note, if needed. 

My hypothesis is overwriting will be more common. When you start a task you might have a general idea of what to do but not the full picture. By the end of the task you can describe exactly what happened.

prat0088:
Thanks for the review!  I’m swamped from work but will try to find time to address the feedback in the coming week.

watson/watson.py
k4nar: Ok for a warning + dropping it :+1: . I guess people will either pass a message to start or stop, but not both.

SoAG:
Any news on this?

joelostblom:
I'm excited about using the functionality in this PR, thanks for all your hard work on it @prat0088!

joelostblom:
@k4nar If this can be merged, I am happy to follow up with a PR addressing your three latest comments:

1. Removing TODO comment.
2. Making the flag plural "notes"
3. Printing the old note before overwriting it.

jmaupetit:
I'll let @prat0088 decide if he can make it or prefer that you finish this awesome feature @joelostblom :pray:

That would make 4 external contributors working on this. You are awesome :sparkles: :heart: :muscle:

prat0088:
Sorry I haven’t been active. Work has changed the way we log time and I no longer have a need for this program, but it is still by far my favorite way I have logged time. 

If this can be merged as is I’ll merge it. If it’s decided changes are needed first then I don’t mind if someone branches this branch, adds the changes, and commits.

jmaupetit:
Thanks for the quick feedback @prat0088!

I would prefer to merge this feature once fully achieved, so @joelostblom feel free to add your fixes to a new branch and open a new PR :pray:

joelostblom:
@jmaupetit  Done! See #383

jmaupetit:
Closing this in favor to #383

PR #340, opened 2019-12-02
Bugfix click completion

MaxG87:
It seems to be more convenient for new users to copy existing completion scripts instead of having to generate them first.

This merge adds the scripts again and updates the relevant documentation.


MaxG87:
The frequent `push --force` flustered me. The two commits I am handing in belated herewith were lost due to a `push --force`, but were asked for explicitly. Please excuse me.

PR #351, opened 2020-01-21
Add optional interactive shell through click-repl

barskern:
By installing `click-repl`[0] the user gets an interactive shell with
tab completion enabled.

[0]: https://github.com/untitaker/click-repl

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -38,6 +38,11 @@ On OS X, the easiest way to install Watson is using [Homebrew](http://brew.sh/):
 ```bash
 $ brew update && brew install watson
 ```
+### Optional interactive shell
+
+If you install [`click-repl`](https://github.com/untitaker/click-repl), watson
+gets a new command called ``repl``, which launches an interactive shell with
+tab-completion.
 
 ### Distribution packages
 
--- a/setup.py
+++ b/setup.py
@@ -47,6 +47,11 @@ setup(
     long_description=readme,
     install_requires=parse_requirements('requirements.txt'),
     tests_require=parse_requirements('requirements-dev.txt'),
+    extras_require={
+        'repl': [
+            'click-repl>=0.1.6',
+        ]
+    },
     entry_points={
         'console_scripts': [
             'watson = watson.__main__:cli',
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -142,6 +142,14 @@ def cli(ctx):
     ctx.obj = create_watson()
 
 
+try:  # pragma: no cover
+    import click_repl
+    click_repl.register_repl(cli)
+    click_repl.register_repl(cli, name="shell")
+except ImportError:
+    pass
+
+
 @cli.command()
 @click.argument('command', required=False)
 @click.pass_context


jmaupetit:
Thank you for this @barskern 

:thinking: I have the feeling that it must be a plugin, and requires therefore that we implement a plugin architecture.

barskern:
Hmm I not sure what you mean. This PR doesn't add anything really besides the fact that it will add a `repl` command if the user already has `click-repl` installed.

jmaupetit:
You've added an extra dependency and its integration is hard coded in the core part of Watson. I think it's not a good architecture, and that we should add plugins support for such feature.

barskern:
I don't mean to come of as being pedantic, but technically I only added metadata about a optional dependency that is not installed with watson. Further if the dependency is not present, there is no change from the previous functionality.

On a broader scale, I do understand that you want to create a plugin system of some sort so that further additions like this can be easily added.

jmaupetit:
> technically I only added metadata about a optional dependency that is not installed with watson. Further if the dependency is not present, there is no change from the previous functionality.

Maybe you don't see it because you are the first one willing to add such pattern, but if we begin to do so, Watson will became a huge bloated mess with optional features and dependencies.

barskern:
Yeah, I understand that point. Hence my previous comment about the fact that I understand that you would like a plugin system regarding other plugins in the future.

I am however not sure how this would be added as a plugin because it needs to modify the `cli` function which is defined in `cli.py`. I will leave this for you to decide on.

davidag:
This PR seems very similar to the one where `did-you-mean` support was added (#318), except in that case the dependency was not optional.

The point to me is if this feature makes sense for Watson users, which I'm not quite sure (I don't see myself using it).

barskern:
Yeah @davidag, it might not fit that well into everyone's workflow, however personally I find it very handy. I have setup a script called `watson-repl` with the following contents:

```bash
#!/usr/bin/env bash
watson report -G
echo "---"
watson status
watson repl
```

And then I have a bind in my `i3` config which launches the script in a floating terminal window:

```
bindsym $mod+Shift+w exec $term_wd --title "watson" --class "floater" -e watson-repl
```

When the script launches I get a terminal with pre-printed status information and a repl where I don't have to start every command with `watson`.

barskern:
Yeah, I understand your reasoning.----My script is very convenient, and I would be happy to make a package for it, however I am not sure of how to do it. Here are some of the options I could consider:- Non-automated "simple" forkIs a lot of work to keep up-to-date.- Automated forkAutomatically pull latest changes from upstream through a webhook and merge into package. Potentially low maintance however probably takes some work to setup correctly.- Simple patch using `diff` and apply before buildWill probably quickly get out-of-date and hence the patch will not apply smoothly.----Any better solutions or tips for my proposed solutions?On 28/01/20 07:56, Yannick PÉROUX wrote:> k4nar commented on this pull request.> > I second @jmaupetit here.> > Adding an optional dependency & 3 lines of conditional code seems harmless, but it means that once this is released, users will expect this option to be supported. We would now depend on the fact that click-repl is maintained & keep working in the future. It doesn't seem actively maintained nor used, so it might not be the best bet.> > I like your example though, it seems very convenient. If you wanted to publish a simple package (watson-repl ;) ) to enable this, we could advertise it in the doc or the readme, as we did with past projects. > > > > -- > You are receiving this because you were mentioned.> Reply to this email directly or view it on GitHub:> https://github.com/TailorDev/Watson/pull/351#pullrequestreview-349490946

barskern:
Yeah, I understand your reasoning.

My script is very convenient, and I would be happy to make a package for it, however I am not sure of how to do it. Here are some of the options I could consider:

* Non-automated "simple" fork

Is a lot of work to keep up-to-date.

* Automated fork

Automatically pull latest changes from upstream through a webhook and merge into package. Potentially low maintance however probably takes some work to setup correctly.

* Simple patch using `diff` and apply before build

Will probably quickly get out-of-date and hence the patch will not apply smoothly.

Any better solutions or tips for my proposed solutions?

k4nar:
Maybe you could package a python script which would include watson, setup the repl, and expose an entrypoint named `watson-repl`.

Or, as @jmaupetit suggested, we could try to setup a plugin system.

PR #357, opened 2020-03-24
Always show total time at bottom of report

### LINKED ISSUE #356: Total time information
I'm looking for a way to obtain total time information of a time period of interest, e.g. the current day.

`watson report --day` outputs a final line about total time of the day's frames, so I could pipe it into `tail -n1`.
However this final line is skipped if only a single project occurred that day (see https://github.com/TailorDev/Watson/blob/master/watson/cli.py#L696 )

Hence the questions:
- Is there a more clever way to obtain total time information of the current day?
- If not; are there objections to remove the condition mentioned above, and always have `watson report` the total time information?

pylipp:
Fix #356

### DIFF
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -691,21 +691,18 @@ def report(watson, current, from_, to, projects, tags, ignore_projects,
                 ))
         _print("")
 
-    # only show total time at the bottom for a project if it is not
-    # an aggregate report and there is greater than 1 project
-    if len(projects) > 1 and not aggregated:
-        _print('Total: {}'.format(
-            style('time', '{}'.format(format_timedelta(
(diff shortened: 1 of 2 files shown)


watson/cli.py
```diff
@@ -691,20 +691,17 @@ def _final_print(lines):
...
         _print("")
 
-    # only show total time at the bottom for a project if it is not
-    # an aggregate report and there is greater than 1 project
-    if len(projects) > 1 and not aggregated:
+    # if this is a report invoked from `aggregate` return the lines; do not
+    # show total time
+    if aggregated:
+        return lines
+    else:
```
jmaupetit: This statement is not required.

PR #362, opened 2020-04-10
Introduce options to configure the ordering of the projects and tags commands

bendardenne:
Introduce `--by {name|time}` and `--reverse` flags for the `projects` and `tags` cli commands.

Sometimes it can be useful to get the recently used projects / tags first. 
My use case is that I have script for [rofi](https://github.com/davatorium/rofi) which allows me to change the watson state interactively. It's convenient for me to have the recently used projects and tags there, an alphabetical list doesn't really help me. I thought other people might benefit from this.

### DIFF
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -681,35 +681,49 @@ def test_pull(mocker, watson):
 # projects
 
 def test_projects(watson):
+    i = 0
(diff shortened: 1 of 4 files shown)


jmaupetit:
Hey @bendardenne! Thx, we'll look into this.

Issue #352, opened 2020-01-22
start --at


prat0088:
After using watson for 3 weeks I've discovered I frequently forget to start timing another task when I switch contexts.  Currently I need to:

1. stop --at
2. start task
3. edit the start time to match the time in 1

I would like:

1. start --at

that would stop the current task as of the --at time, then start a new task as of the --at time... if stop_on_start is set.

davidag:
@prat0088 you can reduce the number of steps using current features:

1. `stop --at`
2. `start -G` (leave no gap)

Nevertheless, your proposal is interesting and I'd probably use it too. Thanks!

prat0088:
Did not realize -G works like that.  Thanks!

hellerbenjamin:
I would use this as well!

sseppola:
Sometimes you forget to start the timer after an intentional gap, `watson start --at` would be great for this scenario.

joelostblom:
I just started using watson and after breaks I keep forgetting to start it until 10-15 min into my task. Maybe I will learn over time, but an `--at` option for `start` would be great!

joelostblom:
For reference, this has also been discussed in issue https://github.com/TailorDev/Watson/issues/75 and PR https://github.com/TailorDev/Watson/pull/224.

jmaupetit:
Hi @joelostblom et al. we heard you loud and clear :wink: Feel free to propose a PR that implements this feature :heart:

joelostblom:
Thanks @jmaupetit ! I created #364.

PR #224, opened 2018-09-15
Specify past or future datetime for frame start or restart with an command line option

braincow:
In issue https://github.com/TailorDev/Watson/issues/75 feature request was made for ability to specify start time for frames that are started and should be considered already running.

This pull request outlines a simple proposal for achieving this. 

Pull request does not implement:
* bash/zsh completion for -f/--from for add and restart commands
* unit tests

PR #364, opened 2020-04-24
Allow specification of start time via `start_at`

### LINKED ISSUE #75: Add an optional (relative) time to the start and stop commands
Some times, you realize you have started working on a task/project without tracking it. It would be nice to have a CLI option to indicate that you actually started earlier. For instance

```
watson start --for 10min 'World Domination Project' +'first analysis'
watson start --since 14:35 'Work on personal organisation'
```

or something similar. 

In the same way, the option can be useful for the `stop` command (e.g. back from lunch and see that your task is still started).

joelostblom:
A suggestion for how to implement `start_at` similar to the already existing `stop_at`. @jmaupetit 

Note that there are 11 tests failing for me on the master branch, which is unchanged with the additions suggested here.

Close  #75 #224 #352

### DIFF
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -635,14 +635,14 @@ def test_pull(mocker, watson):
                 {
                     'id': '1c006c6e-6cc1-4c80-ab22-b51c857c0b06',
                     'project': 'foo',
-                    'start_at': 4003,
+                    'begin_at': 4003,
                     'end_at': 4004,
                     'tags': ['A']
                 },
                 {
                     'id': 'c44aa815-4d77-4a58-bddd-1afa95562141',
                     'project': 'bar',
-                    'start_at': 4004,
+                    'begin_at': 4004,
                     'end_at': 4005,
                     'tags': []
                 }
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -279,13 +279,13 @@ def test_start_project_at(watson):
     watson.start('foo', start_at=now)
     watson.stop()
 
-    # Can't start before the previous task ends
+    # Task can't start before the previous task ends
     with pytest.raises(WatsonError):
         time_str = '1970-01-01T00:00'
         time_obj = arrow.get(time_str)
         watson.start('foo', start_at=time_obj)
 
-    # Can't start in the future
+    # Task can't start in the future
     with pytest.raises(WatsonError):
         time_str = '2999-12-31T23:59'
         time_obj = arrow.get(time_str)
--- a/tests/test_watson.py
+++ b/tests/test_watson.py
@@ -331,13 +331,15 @@ def test_stop_started_project_at(watson):
     watson.start('foo')
     now = arrow.now()
 
+    # Task can't end before it starts
     with pytest.raises(WatsonError):
         time_str = '1970-01-01T00:00'
         time_obj = arrow.get(time_str)
         watson.stop(stop_at=time_obj)
 
-    with pytest.raises(ValueError):
-        time_str = '2999-31-12T23:59'
+    # Task can't end in the future
+    with pytest.raises(WatsonError):
+        time_str = '2999-12-31T23:59'
         time_obj = arrow.get(time_str)
         watson.stop(stop_at=time_obj)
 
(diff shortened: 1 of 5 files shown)


joelostblom:
@jmaupetit  Just added a couple of minor updates. This is ready for review when you get a chance. Really enjoying watson so far and looking forward to contributing more if there is interest! Thanks again for making this great tool!

watson/watson.py
```diff
@@ -251,7 +251,12 @@ def add(self, project, from_date, to_date, tags):
         frame = self.frames.add(project, from_date, to_date, tags=tags)
         return frame
 
-    def start(self, project, tags=None, restart=False, gap=True):
+    def start(self, project, tags=None, restart=False, start_at=None, gap=True):
+        if start_at is not None and not gap:
+            raise WatsonError(
+                'gap=False species the start time to be the end of the '
+                'previoys block, so you cannot use it in conjunction with '
```
jmaupetit, line 258:
```suggestion
                'previous block, so you cannot use it in conjunction with '
```

watson/watson.py
```diff
@@ -251,7 +251,12 @@ def add(self, project, from_date, to_date, tags):
         frame = self.frames.add(project, from_date, to_date, tags=tags)
         return frame
 
-    def start(self, project, tags=None, restart=False, gap=True):
+    def start(self, project, tags=None, restart=False, start_at=None, gap=True):
+        if start_at is not None and not gap:
```
joelostblom, line 255: I noticed now that I could use the `MutuallyExclusiveOption` in the click decorator instead of this if statement. The only drawback would be that I didn't see a way to specify a custom error message which might be helpful, but it would still probably be better to handle this the same as other mutually exclusive options in watson and use the decorator. Let me know what you prefer.

jmaupetit: IMO we should use the decorator when applicable.

joelostblom: Thanks, I changed this to use the decorator in my latest commit. Note the small difference that the previous only raised an error if `--at` was used together with `--no-gap`, while  the click decorator raises and error when used together with either `--gap` or `--no-gap` (it still works with the default parameter value of `gap=True`, it is just not possible to use the flag directly).

joelostblom: Also just a note that there is apparently a click addon for options groups now https://github.com/click-contrib/click-option-group, including mutual exclusiveness and grouping in the help message, might be something worth checking out in the future.

joelostblom:
@k4nar @jmaupetit I restructured my previous changes to fix the failed tests in my latest commit. The only failing test now is related to flake8 and also fails for me on master. Could you review this PR and let me know if you think the changes are sound before I try to add a test specifically for `start_at`?

joelostblom:
I added some suggested tests now also, ready for full review. I am not quite sure the test in `test_cli.py` is necessary, let me know what you think.

tests/test_watson.py
```diff
@@ -311,13 +331,15 @@ def test_stop_started_project_at(watson):
...
+    # Task can't end before it starts
     with pytest.raises(WatsonError):
         time_str = '1970-01-01T00:00'
         time_obj = arrow.get(time_str)
         watson.stop(stop_at=time_obj)
 
-    with pytest.raises(ValueError):
-        time_str = '2999-31-12T23:59'
+    # Task can't end in the future
+    with pytest.raises(WatsonError):
```
joelostblom, line 341: This was previously raisin `ValueError` because there is no month number 31. Now it correctly raises a `WatsonError` if the task starts in the future.

joelostblom:
@jmaupetit @k4nar Just a reminder to review this PR when you get a chance. FWIW I have been using it locally without issues.

jmaupetit:
We'll merge this after #369 :tada:

jmaupetit:
Also: please update the CHANGELOG to add this new feature, yay!

joelostblom:
Thanks @jmaupetit ! I added a note in the changelog and rebased on #369 so should be good to go after that PR is merged.

jmaupetit:
I think you are good to go for a new rebase :trollface: :pray:

joelostblom:
@jmaupetit Ready!

PR #366, opened 2020-04-24
Avoid dubble newlines in csv output

joelostblom:
I noticed that an extra newline character is added to each line of the `--csv` outputs
(displayed as `^M` in vim):

```text
id,start,stop,project,tags^M
76575e4,2020-04-22 19:48:43,2020-04-22 21:15:51,tweaks,optimization^M
128fb31,2020-04-22 21:18:24,2020-04-22 23:19:26,test-proj,"tag1, tag2"^M
```

This is not needed and can cause some programs to accidentally read in a blank line between each line in the csv output:

```text
id,start,stop,project,tags

76575e4,2020-04-22 19:48:43,2020-04-22 21:15:51,tweaks,optimization

128fb31,2020-04-22 21:18:24,2020-04-22 23:19:26,test-proj,"tag1, tag2"
```

This PR is a suggestion for how to fix this and change the output of `--csv` to

```text
id,start,stop,project,tags
76575e4,2020-04-22 19:48:43,2020-04-22 21:15:51,tweaks,optimization
128fb31,2020-04-22 21:18:24,2020-04-22 23:19:26,test-proj,"tag1, tag2"
```

I have only tested this in vim and gedit on Linux.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Always show total time at bottom of report (#356)
+- Use the default system newline character for CSV output.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Always show total time at bottom of report (#356)
-- Use the default system newline character for CSV output.
+- Use the default system newline character for CSV output (#366).
 
 ### Fixed
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -245,14 +245,14 @@ def test_build_csv_empty_data():
 
 
 def test_build_csv_one_col():
-    lt = csv.get_dialect('excel').lineterminator
+    lt = os.linesep
     data = [{'col': 'value'}, {'col': 'another value'}]
     result = lt.join(['col', 'value', 'another value']) + lt
     assert build_csv(data) == result
 
 
 def test_build_csv_multiple_cols():
-    lt = csv.get_dialect('excel').lineterminator
+    lt = os.linesep
     dm = csv.get_dialect('excel').delimiter
     data = [
         co.OrderedDict([('col1', 'value'),
--- a/watson/utils.py
+++ b/watson/utils.py
@@ -358,7 +358,7 @@ def build_csv(entries):
     else:
         return ''
     memfile = StringIO()
-    writer = csv.DictWriter(memfile, header)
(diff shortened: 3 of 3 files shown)


k4nar:
We use the default configuration of `DictWriter`, which is tuned to be close to what Excel expects. Excel being a product from the Windows world, `\r\n` is used instead of just `\n`.

Now changing that would probably break compatibility with quite a few integrations. I think we should keep the default as it looks sane (it's the default of most applications generating CSV as far as I can tell), and trimming the CRLF can be done with Unix tools.

joelostblom:
Thanks for the comment @k4nar. I believe Excel is compatible with `\n` (only have LibreOffice installed to test and it generates `\n` by default), but if there are already integrations relying on the `\r\n` options, then  maybe I could add a config option for this instead?

joelostblom:
I made a suggestion for a configurable `\n` line terminator in my latest commits.

jmaupetit:
I have mixed feelings about this... I think most UNIX users are able to use `dos2unix` or `sed` magic to fix their CSV (if required). Maybe I'm a little biased about this.

joelostblom:
Yes it is true that it is possible to use those tools. I personally think it would be more convenient to have it as an option in the configuration. Do you think there is a drawback with adding it, like not cluttering the config file with too many possible options?

jmaupetit:
> Do you think there is a drawback with adding it, like not cluttering the config file with too many possible options?

Yup.

joelostblom:
Makes sense, what do you think about setting the lineterminator to `os.linesep` instead of adding a config option? That way, users of either platform would get what they expect. This is how the Python data analysis library `pandas` [handles writing of csv files](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_csv.html):

> line_terminator str, optional
> The newline character or character sequence to use in the output file. Defaults to os.linesep, which depends on the OS in which this method is called (‘n’ for linux, ‘rn’ for Windows, i.e.).

jmaupetit:
Good idea! I wonder why this is not the default for the `csv` module.

k4nar:
Ok for me :+1: .

joelostblom:
Great! I added this change and removed the previous ones.

joelostblom:
Thanks @davidag, I updated the two tests that used a lineterminator variable explicitly and added a note in the changelog.

jmaupetit:
Hey  @joelostblom can you rebase your branch so that I can merge this. Thanks!

joelostblom:
@jmaupetit Rebased!

jmaupetit:
:thinking: I guess a new arrow release is breaking the build... again. :disappointed:

joelostblom:
Yes, i submitted another PR to pin the arrow version https://github.com/TailorDev/Watson/pull/372

jmaupetit:
Yup saw it, merged it! Ready for another rebase?

joelostblom:
Done!

PR #369, opened 2020-04-25
Add option to control the order of the days in the `log` output

joelostblom:
I think it is quite common to sort with the newest entry at the bottom (e.g with `ls` and this is also how the times are ordering within each day in the `log` output), so I suggest adding an option to enable this behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Log output order can now be controlled via the `--reverse/--no-reverse` flag
+  and the `reverse_log` configuration option (#369)
+
 ### Changed
 
 - Require latest Arrow version 0.15.6 to support ISO week dates (#380)
@@ -17,8 +22,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Improve Arrow 0.15.0 support after changes in `arrow.get()` behavior (#296)
 - Watson now suggests correct command if users make small typo (#318)
-- Log output order can now be controlled via the `--reverse/--no-reverse` flag
-  and the `reverse_log` configuration option (#369)
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


joelostblom:
Will add a flag with the same name `--log_latest_day_on_top`!

joelostblom:
That's what I had initially, but then I wanted to be more specific to indicate that it is only reversing the order of the days, not the time entries within the days, so that I would leave it open for someone to add an option to reverse times if desired. This might never be the case though, and I am happy to rename both the option and the flag `reverse-log` if you prefer that.

jmaupetit:
`--reverse` should be fine, thanks!

joelostblom:
I can try, but might need some assistance to make it satisfactory. Before giving it a go, I just want to clarify what what you would like this test to contain. Do you want it to be something simple like the [current CLI log tests](https://github.com/TailorDev/Watson/blob/af8911715ce131218c413751197b6f0ce1882b5b/tests/test_cli.py#L131-L150), just adding the `--reverse` flag and check that no errors are raised? Or do you want there to be a new `log` section in `test_watson.py` which checks the actual ordering of the days when using `--reverse/--no-reverse`?

jmaupetit:
Testing Watson's API is easier to implement, so adding a new log section in `test_watson.py` would be perfect.

joelostblom:
@jmaupetit  Let me know what you think of the tests I just added. Since `log` is defined entirely in `cli.py`, I could not figure out a way to test it in `watson.py`. I tried adding it in with the other cli-tests but could not figure out a good way to add data and test it, so I added the tests to `test_utils.py`. The disadvantage is that we're really only testing the inner most function and things could of course go wrong outside, but I wasn't able to implement anything better. Let me know if you have suggestions. The failing flake8test is unrelated and fails on master as well.

joelostblom:
@jmaupetit Just a reminder to have a look at this when you get a chance.

jmaupetit:
Thanks @joelostblom ! If you handle the rebase, I'll merge this asap :pray:

joelostblom:
Haha thanks! =) I just rebased and squashed a few of the commits.

joelostblom:
Realized that the changelog entry was in the wrong place since the 1.9 tag was added recently. Ready for merge now :+1:

PR #376, opened 2020-05-22
[Requirements] add colorama for Windows

MinchinWeb:
Click seems to have built-in color support, but this only kicks on Windows if `colorama` is also installed.

The PR adds colorama to the requirements for Windows only to turn on color support.

### DIFF
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -1047,7 +1047,6 @@ def log(watson, current, from_, to, projects, tags, year, month, week, day,
                 project=style('project', u'{:>{}}'.format(
                     frame.project, longest_project
                 )),
-                pad=longest_project,
                 tags=(" "*2 if frame.tags else "") + style('tags', frame.tags),
                 start=style('time', '{:HH:mm}'.format(frame.start)),
                 stop=style('time', '{:HH:mm}'.format(frame.stop)),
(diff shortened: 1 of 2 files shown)


watson/cli.py
```diff
@@ -1047,7 +1047,6 @@ def _final_print(lines):
                 project=style('project', u'{:>{}}'.format(
                     frame.project, longest_project
                 )),
-                pad=longest_project,
```
jmaupetit, line 1050: Can explain this line removal a bit more?

MinchinWeb: Travis wouldn't pass with it in place. See [this run](https://travis-ci.org/github/TailorDev/Watson/jobs/689886199). This is the actual error message:

```
flake8 run-test: commands[0] | flake8 --show-source watson/ tests/ scripts/

watson/cli.py:1045:13: F522 '...'.format(...) has unused named argument(s): pad
```

MinchinWeb: (I'm happy to add it back, if that's what you think should happen)

jmaupetit: No, no, you are totally right (and so does flake8). I'm surprised that we did not see it before.

jmaupetit:
Thanks @MinchinWeb :pray: We are lucky to have you for Windows-related improvements :heart:

MinchinWeb:
Thank you for your kind words!

PR #380, opened 2020-06-07
Fix tests for arrow>=0.15.6

geistesk:
Since its last release[0], the heavily used arrow library supports ISO
week dates[1]. The CLI tests assumed that arrow does not support this.
However, the new version nullified this assumption.

As a provisional measure, the version of arrow was limited upwards in
PR #372[2]. An obsolete version was thereby requested.

A current version, or at least support for one, is important for
third-party package managers. Especially GNU/Linux distributions prefer
to use their own package manager to install software over pip.

Thus, this commit removes both the restriction to an outdated arrow
version in the requirements.txt and validates previously invalid marked
week dates.

[0]:https://github.com/crsmithdev/arrow/releases/tag/0.15.6
[1]:https://en.wikipedia.org/wiki/ISO_week_date
[2]:https://github.com/TailorDev/Watson/pull/372


jmaupetit:
It's a strong requirement, but I've no hard-feelings about this. @k4nar & @joelostblom WDYT?

geistesk:
Currently I see three options:

1. Enforce an outdated version of `arrow` and keep the old tests.
2. Enforce the latest version of `arrow` and use the updated tests.
3. Remove the ISO week date-test and a limitation to `arrow`'s version.

Imho, removing tests is never good. Thus, one has to pick a side of `arrow`'s _version barrier_: use an old version or force the (currently) newest.

joelostblom:
I don't have anything against this either. I am not very experienced with what Linux distros prefer so I defer judgement there (although my general problem with many have been that they have too old packages so not sure if they would have such a recent version of arrow).

One additional alternative could be to have a separate version requirement on arrow for development, since it only breaks the tests and list in the docs that week dates are only available on the latest version of arrow, which would avoid limiting the availability of Watson to those that don't care about week dates or running the tests, but for some reason can't get the latest arrow (maybe this is not at all common and therefore moot).

Just throwing out ideas.

geistesk:
I created this PR as a contributor to a distribution - NixOS, if it matters. The latest `arrow` version thus prevented me from version bumping watson. As options I saw adding a custom patch, submitting a patch upstream or using an extra `arrow` package only for watson.

Others are facing the same problem, e.g., [openSUSE have patched watson](https://build.opensuse.org/package/view_file/openSUSE:Factory/watson/watson-arrow-version.patch?expand=1) to remove `arrow`'s version limitation. Nevertheless, it's dark out there regarding packages - both for [`arrow`](https://repology.org/project/python:arrow/versions) and [watson](https://repology.org/project/watson/versions).

Yet enforcing an old version leads to incompatibilities in the long run. Incentives for updating should therefore be provided.

jmaupetit:
@geistesk you convinced me :wink: Waiting for @k4nar opinion on this.

jmaupetit:
Ok, then. @geistesk can you update the changelog so that I can merge this. Thank you! :pray:

geistesk:
@jmaupetit: Does it look right to you now?

PR #382, opened 2020-06-16
Add after edit check that start is in the past, closes #381

### LINKED ISSUE #381: After edit check doesn't check that the starting time is in the past
I just recent made an error while editing an event, where I accidentally set the starting time in the future. I did however not get an error at that time, only when I tried to stop the event did watson complain with:

```
Error: Task cannot end before it starts.
```

It seems like it could be caught already in the after edit-check.

I'm using watson 1.8.0

KennethNielsen:
This PR adds an after edit check that checks whether the start time is in the past. The PR was triggered by myself accidentally doing this during an edit of the current task and noticing that I didn't get an error before stopping the task, which is mildly confusing.

However. This PR raises another question. Should we be doing this kind of check at all, and what really is the policy. Is the end time allowed to be in the future? I Noticed e.g. that add doesn't seem to be doing of these sanity checks. Should it?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Require latest Arrow version 0.15.6 to support ISO week dates (#380)
 
+### Fixed
+
+- Make after-edit-check ensure that edited time stamps are not in the future
+  (#381)
+
+
 ## [1.9.0] - 2020-05-27
 
 ### Added
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -1297,6 +1297,8 @@ def edit(watson, confirm_new_project, confirm_new_tag, id):
                     "Task cannot end before it starts.")
             if start > arrow.utcnow():
                 raise ValueError("Start time cannot be in the future")
+            if stop and stop > arrow.utcnow():
+                raise ValueError("Stop time cannot be in the future")
             # break out of while loop and continue execution of
             #  the edit function normally
             break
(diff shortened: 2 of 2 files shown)


KennethNielsen:
This PR closes #381

watson/cli.py
```diff
@@ -1269,6 +1269,8 @@ def edit(watson, confirm_new_project, confirm_new_tag, id):
             if not watson.is_started and start > stop:
                 raise ValueError(
                     "Task cannot end before it starts.")
+            if start > arrow.utcnow():
+                raise ValueError("Start time cannot be in the future")
```
jmaupetit, line 1273: Maybe we should also check that the stop time is not in the future?

KennethNielsen: Right. I will add that and update the changelog, forgot about that.

jmaupetit:
LGTM! 

@KennethNielsen I've tried to update your branch from GitHub, but the CHANGELOG is messy. Can you rebase this onto master so that I can merge this asap? :pray:

KennethNielsen:
It's rebased.

CHANGELOG.md
```diff
@@ -37,6 +37,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Aggregate ignores frames that crosses aggregate boundary (#248)
 - The `report` and `aggregate` commands with `--json` option now correctly
   encode Arrow objects (#329)
+- Make after-edit-check ensure that edited time stamps are not in the future
+  (#381)
```
jmaupetit, line 41: Should be in the `fixed` subsection of the `Unreleased` section.

KennethNielsen: Apologies. Must have been too tired. Should be fixed now.

PR #383, opened 2020-06-18
Add --note flag to associate messages with frames

joelostblom:
@jmaupetit This is a continuation of #337. As mentioned there, I merged master into this branch because the rebase was a headache with so many conflicting changes (possible I'm not doing this optimally). I suspect this doesn't matter since you usually squash when merging, but it would be great if this was given priority over other PR's, since it is hard to rebase if master is updated.

I added @k4nar 's suggestions in the three first commits. Just a comment that although the plural form `--notes` is more natural for log since it is short for `short_notes`, it might be confusing since it is `--note` everywhere else. I'll let you decide what to do, if anything.

I added three commits of my own, only 4e0e4b6 is a functional change. Without it, I would not be able to edit frames since the new `note` field would be interpreted as the `updated_at` field and `watson edit` raised an error trying to convert this to a date. I have added this and a few more comments on the code also.

Close #337 #252 #93


watson/cli.py
```diff
@@ -1067,19 +1125,22 @@ def _final_print(lines):
...
-                    frame.project, longest_project
-                )),
-                tags=(" "*2 if frame.tags else "") + style('tags', frame.tags),
-                start=style('time', '{:HH:mm}'.format(frame.start)),
-                stop=style('time', '{:HH:mm}'.format(frame.stop)),
-                id=style('short_id', frame.id)
-            )
-            for frame in frames
-        ))
+        for frame in frames:
```
joelostblom, line 1128: I was not sure how to incorporate the if-statement in this section, so this is the only place in the merge where I went with the older syntax with the traditional for loop. Let me know if you prefer the have this change reverse and how it would be written with the 'note' if statement below.

watson/utils.py
```diff
@@ -420,3 +423,10 @@ def json_arrow_encoder(obj):
         return obj.for_json()
 
     raise TypeError("Object {} is not JSON serializable".format(obj))
+
+
+def format_note(note):
+    return u"{}{}".format(
+        style('note', '>> '),
+        style('note', note.replace('\n', '\n' + ' '*20))
```
joelostblom, line 431: This aligns rows of notes containing `\n` in the log output.

watson/watson.py
```diff
@@ -302,9 +305,16 @@ def stop(self, stop_at=None):
         if stop_at > arrow.now():
             raise WatsonError('Task cannot end in the future.')
 
+        if note is None:
+            note = old.get('note')
+        elif old.get('note') is not None:
+            print('Overwriting old note:\n>> {}'.format(old.get('note')))
```
joelostblom, line 311: This means that this warning is printed first, before the always printed watson stop message. This is a bit visually jarring, but it might be a good thing since it is an important thing to note, see screenshot. I tried passing a variable up to watson cli and then use echo after the stop message, but could not find a good way of doing this since the frameattributesare actually properties are note modifieable so I could not pass say a tuple and then print the old note and overwrite the property with the new note.

image

joelostblom:
@k4nar Just a reminder to have a look at this when you get a chance.

joelostblom:
@k4nar and @jmaupetit I have been running this branch locally since opening the PR and it seems to be working well (although I don't use notes that often tbh). Just a reminder to review this PR when you have time.

aero31aero (approved):
Just a reminder, this is a very useful feature imo and should be merged. I looked at the code and it looks fine to me. Tested the branch locally and it works as expected.

Except for maybe the rebase artifacts here, this LGTM.

joelostblom:
Just a note that I didn't use this feature much and I am no longer running this branch locally. I ran into some issues with my frames file format when I moved it to a new watson installed from the main branch, so I went ahead and deleted the notes fields for each frame and it is now working again. I didn't have time to investigate exactly why this issue occurred, but thought I would leave a note that  there might be incompatibilities with newer watson versions in case others are considering running from this PR (you probably should wait until when/if this feature is merged into main).

PR #388, opened 2020-07-11
Add day and week editing options

Link-:
A useful feature for daily Watson users (like myself) is having the ability to edit all the entries of a given day or week (generally towards the end of the day | week). This PR provides 5 new optional flags to the `edit` command:

```
`-d, --day` | Edit all frames for today.
`-w, --week` | Edit all frames for the past week.
`-m, --month` | Edit all frames for the past month.
`-f, --from` | The date from when the log should start.
`-t, --to` | The date at which the log should stop (inclusive).
```

If day, week or month are passed, all the frames for that day, week or month will be available for editing. If an id is also passed it will be ignored.

If from and/or to are specified the frames within this range will be available for editing.

**Notes:**
- [x] Unit tests have been included for the newly added utility functions
- [x] Documentation has been updated to reflect the new flags
- [x] API was maintained and is consistent with what's released right now

### DIFF
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -36,6 +36,7 @@ from .utils import (
     get_start_time_for_period,
     get_frames_for_today,
     get_frames_for_week,
+    get_frames_for_month,
     options, safe_save,
     sorted_groupby,
     style,
@@ -1225,10 +1226,12 @@ def add(watson, args, from_, to, confirm_new_project, confirm_new_tag):
               help="Edit all frames for today.")
 @click.option('-w', '--week', is_flag=True, default=False,
               help="Edit all frames for the past week.")
+@click.option('-m', '--month', is_flag=True, default=False,
+              help="Edit all frames for the past month.")
 @click.argument('id', required=False, autocompletion=get_frames)
 @click.pass_obj
 @catch_watson_error
-def edit(watson, confirm_new_project, confirm_new_tag, day, week, id):
+def edit(watson, confirm_new_project, confirm_new_tag, day, week, month, id):
     """
     Edit one or more frames.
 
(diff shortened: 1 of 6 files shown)


watson/cli.py
```diff
@@ -1242,28 +1251,40 @@ def edit(watson, confirm_new_project, confirm_new_tag, id):
...
-        'start': frame.start.format(datetime_format),
-        'project': frame.project,
-        'tags': frame.tags,
-    }
-
-    if id:
-        data['stop'] = frame.stop.format(datetime_format)
+    data = [
+        {
+            'id': frame.id,
```
jmaupetit, line 1279: This will break if we edit the current frame that has no id yet?

jmaupetit:
Thx for this clean implementation! We'll look in more details quickly. :pray:

loonies:
Just to be consitent - what about other time slicing options: `--month`, `--year`, `--from`, and `--to`?

watson/cli.py
Link-: Thank you for your review @jmaupetit. No this will not break as the id will have a value of `null` as follows:

```
[
    {
        "id": null,
        "project": "NoId",
        "start": "2020-07-16 21:02:09",
        "stop": "2020-07-16 21:02:12",
        "tags": [
            "Frame with no id yet"
        ]
    }
]
```

in this implementation, `id`s are immutable, meaning even if the user tries to change them they will be reverted back to original values. In this case `null`.

Link-:
> Just to be consitent - what about other time slicing options: `--month`, `--year`, `--from`, and `--to`?

Definitely feasible and simple to implement. I'll add them to this PR 👍

jmaupetit:
> I was thinking of adding them to this PR but the branch name becomes inconsistent with implementation then.

We don't really care about your branch name, you can pursue your work in here. It will be squashed before merging. Thx again!

Link-:
Apologies for the delay in finalising this implementation but I couldn't find the free time during the past couple of weeks. This implementation is now complete with `from` and `to` options implemented. I've also made some updates to the response after successful editing: instead of returning the whole list of frames in the specified range now it only returns the frames that were actually edited only.

jmaupetit:
Please don't apologies for having better things to do. We are all volunteers here :wink:

Link-:
Small bump to this PR; anything else is required before merging this?

jmaupetit:
I should take time to test it locally...

Link-:
@jmaupetit I'll just repeat what you told me: "Please don't apologies for having better things to do. We are all volunteers here"  :D I was just curious as to the next steps that's all, no pressure!

Link-:
This PR is now broken

Issue #392, opened 2020-09-08
Question / new feature: Mark frames as "billed"


gnanet:
The fetaure described below would help billing of work that has been done:

1.  if there would be a possibility to mark frames as billed, by specify a from+to+project+tags search, or specify a frame-ID directly

1. and a switch for filtering frames which are already billed in the log, and report functions.

These together would allow the report and log functions show and count only non billed frames.

jmaupetit:
Wouldn't a `billed` tag be relevant here? Billed frames can be edited to add such tag.

gnanet:
Maybe i did not look enough at the command switsches, but i have not found for the commands log and report any switch to "exclude" specific tags. If there is  a way to do that, i would say, we document the way how this kind of billing-assistive usage works, and we are done.

Other way would be to "move" billed frames to a project, that is named like "origproject-billed"

jmaupetit:
The `report` command has a `--ignore-tag TAG` flag, but unfortunately the `log` command doesn't :disappointed:

gnanet:
No, that is not an unfortunate thing, because then i have the example code what i (hopefully) simply can replicate, and that is a good info to start with. I will get home on monday and i can start try, and place a PR on the result.

jmaupetit:
Oh, wow, thanks! :pray:

gnanet:
As i see, cli->report uses the watson.report function, which relies on the same frames.filter function than it is used in the cli->log . so i already added all relevant parts from report to log in cli.py, and distilled the validation and exception for projects vs ignored_projects; and tags vs ignored_tags. First checks looking good.

Next i have to create an add / remove tag from tags by key function pair, using something similar


```python
    def add_tag_to_frame(self, key, tag_to_add):
        """Add a tag to frame by key."""

        updated_at = arrow.utcnow()
        # add tag

        self.frames[fkey] = self.frames[key]._replace(
                            tags= self.frames[key].tags + tag_to_add,
                            updated_at=updated_at
                        )

    def remove_tag_from_frame(self, key, tag_to_remove):
        """Remove a tag from frame by key."""

        updated_at = arrow.utcnow()
        # remove tag

        self.frames[key] = self.frames[key]._replace(
                            tags= self.frames[key].tags.remove(tag_to_remove),
                            updated_at=updated_at
                        )

```

gnanet:
the change of cli.py if you would like to have a look at it

```diff
--- cli.py	2020-07-04 02:11:43.258650000 +0200
[...]

PR #394, opened 2020-09-15
New feature: addtag / removetag cli commands

gnanet:
The changes of this PR are directly related to #392

The two new commands allow the addition/removal of a single tag (default="billed") to a frame specified by ID, which helps to mark tasks that are already billed to the client.

In combination with the ignore parameters available for both watson log and watson report, already billed timeframes can be hidden, to aid the collection of non-billed work-items.

### DIFF
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -919,6 +919,14 @@ def aggregate(ctx, watson, current, from_, to, projects, tags, output_format,
               help="Logs activity only for frames containing the given "
               "tag. You can add several tags by using this option multiple "
               "times")
+@click.option('--ignore-project', 'ignore_projects', multiple=True,
+              help="Reports activity for all projects but the given ones. You "
+              "can ignore several projects by using the option multiple "
+              "times. Any given project will be ignored")
+@click.option('--ignore-tag', 'ignore_tags', multiple=True,
+              help="Reports activity for all tags but the given ones. You can "
+              "ignore several tags by using the option multiple times. Any "
+              "given tag will be ignored")
 @click.option('-j', '--json', 'output_format', cls=MutuallyExclusiveOption,
               flag_value='json', mutually_exclusive=['csv'],
               multiple=True,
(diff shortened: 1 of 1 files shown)


gnanet:
tried the squashing of all small-fix commits into one singe commit,but im am not sure if that went as expected ...

jmaupetit:
:thinking: I am not sure that we need a new command to add or remove tags. IMO the `edit` command should be used.

gnanet:
The edit command, as i understand it, opens the frame's "source" in a texteditor, and my goal was to "help billing of work that has been done"

Short explanation:

If i do my monthly billing, i go trough the watson log for every project, and collect the tasks(frames) which can be included into the invoice. But after the second month, i had already problems to distinct billed and non-billed work.

The add/remove tag command allows me, to run a small bash loop, where the frame ids (grepped from watson log --from XX --to YY -p project --ignore-tag billed )  to include in my bill, and get them quickly marked as billed.

If i would have to go trough 10-30 frames "by hand" editing their JSON source, the time of administration would explode.

BUT, i can imagine to extend the edit command to make these single-tag changes upon some switches and parameters, without the need to open the texteditor.

All of all, this is only a proof-of-concept, where i learned a lot, and the PR was only created because of this comment: https://github.com/TailorDev/Watson/issues/392#issuecomment-690912022

jmaupetit:
> BUT, i can imagine to extend the edit command to make these single-tag changes upon some switches and parameters, without the need to open the text editor.

Yes! :100: yes!

gnanet:
"this is only a proof-of-concept, where i learned a lot"

So you want me to learn more...

gnanet:
Here's the plan:

1. i have to cancel this PR
2. then create a new PR separately for the --ignore... options added to the log command
3. do some play with trial-and-error to extend edit command
4. create a new PR with the extended edit code

gnanet:
Cancelinng this PR in favor of the plan described above

jmaupetit:
> So you want me to learn more...

:sweat_smile: I have no idea of what could be your technical background, but it seems that you are a fast-learner :muscle:

Issue #407, opened 2021-02-27
Breaking changes with arrow 1.0.0


veganjay:
## Problem Description

I installed *td-watson* in a fresh environment and received an error when running:

```
$ docker run -it --rm python:latest /bin/bash
root@4156557dc8cf:/# pip install td-watson
...
root@4156557dc8cf:/# watson status
Traceback (most recent call last):
  File "/usr/local/bin/watson", line 5, in <module>
    from watson.__main__ import cli
  File "/usr/local/lib/python3.9/site-packages/watson/__main__.py", line 2, in <module>
    from watson import cli
  File "/usr/local/lib/python3.9/site-packages/watson/cli.py", line 441, in <module>
    _SHORTCUT_OPTIONS_VALUES = {
  File "/usr/local/lib/python3.9/site-packages/watson/cli.py", line 442, in <dictcomp>
    k: get_start_time_for_period(k) for k in _SHORTCUT_OPTIONS
  File "/usr/local/lib/python3.9/site-packages/watson/utils.py", line 193, in get_start_time_for_period
    start_time = get_last_full_moon(now)
  File "/usr/local/lib/python3.9/site-packages/watson/fullmoon.py", line 226, in get_last_full_moon
    idx = bisect.bisect_right(fullmoons, now)
TypeError: '<' not supported between instances of 'method' and 'int'
```

The *arrow library* was [updated on Feb 24, 2021](https://arrow.readthedocs.io/en/stable/releases.html) and there are [breaking changes](https://github.com/arrow-py/arrow/issues/832).

The default *td-watson* installation uses the new version,  *arrow 1.0.0*:

```
root@4156557dc8cf:/# pip freeze
arrow==1.0.0
certifi==2020.12.5
chardet==4.0.0
click==7.1.2
click-didyoumean==0.0.3
idna==2.10
python-dateutil==2.8.1
requests==2.25.1
six==1.15.0
td-watson==1.10.0
urllib3==1.26.3
```

## Workaround

Downgrade to *arrow 0.17.0*:

```
root@4156557dc8cf:/# pip uninstall arrow
Found existing installation: arrow 1.0.0
Uninstalling arrow-1.0.0:
  Would remove:
    /usr/local/lib/python3.9/site-packages/arrow-1.0.0.dist-info/*
    /usr/local/lib/python3.9/site-packages/arrow/*
Proceed (y/n)? y
  Successfully uninstalled arrow-1.0.0
[...]

PR #408, opened 2021-03-01
Issue #407: Migrate to arrow 1.0.0

veganjay:
The arrow library moved to 1.0.0 on February 24, 2021 and with it came breaking changes.
This pull request addresses those changes per the [arrow migration guide](https://github.com/arrow-py/arrow/issues/832).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-arrow>=0.15.6
+arrow==1.0.0
 click>=7.0
 click-didyoumean
 colorama; sys_platform == "win32"
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-arrow==1.0.0
+arrow>=1.0.0
 click>=7.0
 click-didyoumean
 colorama; sys_platform == "win32"
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,6 @@ matrix:
           os: linux
           dist: xenial
           env: TOXENV=flake8
-        - python: 3.5
-          os: linux
-          dist: trusty
-          env: TOXENV=py35
         - python: 3.6
           os: linux
           dist: trusty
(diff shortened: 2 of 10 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
-arrow>=0.15.6
+arrow==1.0.0
```
jmaupetit, line 1: Let's be optimistic!

```suggestion
arrow>=1.0.0
```

jmaupetit: I hope this won't have a catastrophic impact for people installing Watson in their global python distribution...

jmaupetit (commented):
Looks like you'll have to also fix the tests... Thanks! :pray:

veganjay:
> Looks like you'll have to also fix the tests... Thanks!

I took a quick look and there is an issue with timezone localization, making "add --from" and "add --to" work incorrectly.  I need to investigate further.

veganjay:
I fixed the unit test cases and also another issue with the timezone.  The proper way to change the timezone of an arrow date is now:

    date = date.replace(tzinfo=tz.tzlocal())

The travis build failed because arrow 1.0.0 is not available for python 3.5.
Given that [Python 3.5 is at end of life](https://www.python.org/downloads/release/python-3510/), it may be worthwhile considering ending support with updating setup.py and .travis.yml accordingly.

jmaupetit:
Thank you so much @veganjay :pray: 

>  Given that Python 3.5 is at end of life, it may be worthwhile considering ending support with updating setup.py and .travis.yml accordingly.

Agreed! Can you add consequent changes in this PR & fill the CHANGELOG to mention arrow major release upgrade + dropped python 3.5 support? Thx :heart:

veganjay:
Removed support for 3.5 and updated the CHANGELOG.

By the way, I installed *td-watson* with [pipx](https://github.com/pipxproject/pipx):

    pipx install td-watson

This way the arrow library and other dependencies are installed in their own isolated environment.  This is helpful if working on various other virtualenvs during the day.

jmaupetit (approved):
Thank you @veganjay :pray: :heart:

Issue #430, opened 2021-05-12
Crash caused by ValueError


Sloff:
Hi, I'm receiving the following when I try to run Watson on my Windows 10 pc:

```
Traceback (most recent call last):
  File "c:\users\ettienne.pitts\appdata\local\programs\python\python39\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\users\ettienne.pitts\appdata\local\programs\python\python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\ettienne.pitts\AppData\Local\Programs\Python\Python39\Scripts\watson.exe\__main__.py", line 4, in <module>
  File "c:\users\ettienne.pitts\appdata\local\programs\python\python39\lib\site-packages\watson\__main__.py", line 1, in <module>
    from watson import cli
  File "c:\users\ettienne.pitts\appdata\local\programs\python\python39\lib\site-packages\watson\cli.py", line 529, in <module>
    def report(watson, current, from_, to, projects, tags, ignore_projects,
  File "c:\users\ettienne.pitts\appdata\local\programs\python\python39\lib\site-packages\click\decorators.py", line 247, in decorator
    _param_memo(f, OptionClass(param_decls, **option_attrs))
  File "c:\users\ettienne.pitts\appdata\local\programs\python\python39\lib\site-packages\watson\cli.py", line 46, in __init__
    super(MutuallyExclusiveOption, self).__init__(*args, **kwargs)
  File "c:\users\ettienne.pitts\appdata\local\programs\python\python39\lib\site-packages\click\core.py", line 2467, in __init__
    super().__init__(param_decls, type=type, multiple=multiple, **attrs)
  File "c:\users\ettienne.pitts\appdata\local\programs\python\python39\lib\site-packages\click\core.py", line 2106, in __init__
    raise ValueError(
ValueError: 'default' must be a list when 'multiple' is true.
```

Wasper256:
Got the same issue on Ubuntu 20.04LTS. Looks like it happened after python version update(Python 3.8.5) or some other updates in system. 

As a temporary solution I've reinstalled tool on Python 3.6 and it works for me now.

jmaupetit:
Looks like a Click release incompatibility. Which version of Watson are you running? Have you tried to uninstall/reinstall it?

Sloff:
I did a pip install earlier today, so I assume the latest version? How can I confirm?

[...]

PR #431, opened 2021-05-12
Fix click compatibility range

jmaupetit:
### Purpose

Watson is not compatible with Click 8+. See #430 

### Proposal

- [x] Update click compatibility range


k4nar (approved):
Ok as a temporary solution :) .

PR #432, opened 2021-05-20
Update click to 8

DavidOliver:
**Warning:** I don't know what I'm doing; I don't usually use Python and this is my first time with Watson's code, so this pull request should be treated with caution. But hopefully it can at least act as the start of a fix for use of click v8.

The cause of #430 seems to be this ([click v8 changes](https://click.palletsprojects.com/en/8.0.x/changes/#version-8-0-0)):

> TypeError is raised when parameter with multiple=True or nargs > 1 has non-iterable default. [#1749](https://github.com/pallets/click/issues/1749)

['Multiple options' at click documentation](https://click.palletsprojects.com/en/8.0.x/options/#multiple-options).

I remove `multiple=True` from several click options, which I don't believe are intended to be specified at the command line more than once as they seem to be flags rather than options which take (multiple) values.

I believe I've successfully run Watson's tests, which all seem to be passing.

By the way, will a new release of Watson be prepared soon, either with the temporary fix of specifying v7 of click or this fix? I ask as I'm currently working on my system config. :)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 arrow>=1.0.0
-click>=8.0,<9.0
+click>=8.0
 click-didyoumean
 colorama; sys_platform == "win32"
 requests
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 arrow>=1.0.0
-click>=7.0,<8.0
+click>=8.0,<9.0
 click-didyoumean
 colorama; sys_platform == "win32"
 requests
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -204,7 +204,7 @@ def _start(watson, project, tags, restart=False, start_at=None, gap=True):
               help=("(Don't) leave gap between end time of previous project "
                     "and start time of the current."))
 @click.argument('args', nargs=-1,
-                autocompletion=get_project_or_task_completion)
+                shell_complete=get_project_or_task_completion)
 @click.option('-c', '--confirm-new-project', is_flag=True, default=False,
               help="Confirm addition of new project.")
 @click.option('-b', '--confirm-new-tag', is_flag=True, default=False,
@@ -316,7 +316,7 @@ def stop(watson, at_):
                     '(YYYY-MM-DDT)?HH:MM(:SS)? format.'))
 @click.option('-s/-S', '--stop/--no-stop', 'stop_', default=None,
               help="(Don't) Stop an already running project.")
-@click.argument('frame', default='-1', autocompletion=get_frames)
+@click.argument('frame', default='-1', shell_complete=get_frames)
 @click.pass_obj
 @click.pass_context
 @catch_watson_error
@@ -494,11 +494,11 @@ _SHORTCUT_OPTIONS_VALUES = {
               flag_value=_SHORTCUT_OPTIONS_VALUES['all'],
               mutually_exclusive=['day', 'week', 'month', 'luna', 'year'],
               help='Reports all activities.')
-@click.option('-p', '--project', 'projects', autocompletion=get_projects,
+@click.option('-p', '--project', 'projects', shell_complete=get_projects,
               multiple=True,
               help="Reports activity only for the given project. You can add "
               "other projects by using this option several times.")
-@click.option('-T', '--tag', 'tags', autocompletion=get_tags, multiple=True,
+@click.option('-T', '--tag', 'tags', shell_complete=get_tags, multiple=True,
               help="Reports activity only for frames containing the given "
               "tag. You can add several tags by using this option multiple "
               "times")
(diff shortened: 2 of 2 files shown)


jmaupetit:
Wow this is huge. Thanks :pray: We'll look into this soon.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 arrow>=1.0.0
-click>=7.0,<8.0
+click>=8.0,<9.0
```
k4nar, line 2: I don't think we should explicitly forbid 9.0, we don't know if that will be a breaking change for us yet. We usually try to avoid pinning versions if we can.

k4nar (commented):
Indeed I don't think those should have been marked as "multiple", this was probably a bad copy/paste in the first place :smile: .

This looks to me, except for the small comment about the pinned version :+1: . Thank you very much!

requirements.txt
DavidOliver: I restricted to 8 partly because it was a "MAJOR" version change (7 to 8) that broke Watson. I'm not familiar with either project, but shouldn't major releases of dependencies only be (explicitly) updated to once they've been tested due to the likelihood of breaking changes?

k4nar: This is a long debate in the Python community :) . So far my opinion is to allow any version of a dependency as long as you don't know for sure that you're not compatible with it. This is less of an issue if you isolate the dependencies of each Python project you're using, but if you're not doing it things can get messy pretty quickly if everybody is asking for specific versions of a package for no proper reason.

DavidOliver: Huh - interesting! Seems like asking for trouble to me. :D Will you remove the <9 requirement or should I?

mizhozan:
I'm not pro Python user and don't much about test suits, but my test with this pr on my local computer works find plus I did a simple pytest and there was 250 passed, 14 warning, 91 errors, you can see the result [here](https://gist.github.com/mizhozan/abbec077175c17c95687d9717f2c6f16).

DavidOliver:
@mizhozan, I also had those fixture-related failures. The solution seemed to be install the dev requirements from [requirements-dev.txt](https://github.com/TailorDev/Watson/blob/master/requirements-dev.txt). (See the varying mock versions depending on the version of setuptools you have.)

mizhozan:
Thanks a lot, now the test runs perfectly with 341 passes and 14 warnings [accordingly](https://gist.github.com/mizhozan/a77a5e0b5afd68ac10370eb4a95b5ad0) in my case.at 

Is there anything hinder merging this PR to the main release? I like to have watson back (currently I've switched to timew, temporarily) and don't want to mess with the OS python site packages.

> @mizhozan, I also had those fixture-related failures. The solution seemed to be install the dev requirements from [requirements-dev.txt](https://github.com/TailorDev/Watson/blob/master/requirements-dev.txt). (See the varying mock versions depending on the version of setuptools you have.)

requirements.txt
jmaupetit: Hi David! Can you fix this, and then I'll merge your work and release it. Thanks!

DavidOliver: Thanks! Done.

DavidOliver:
@mizhozan Thanks. I think I managed to confuse myself regarding the versions of packages I had installed. I've just (re-?)re-installed click (8.0.1), and now I also have the deprection warnings.

@jmaupetit Can you hold off on merging, please? I'll see if I can get rid of the deprecation warnings.

DavidOliver:
The deprecations warnings are now gone (see last commit), but I'm out of my depth here and can't know whether or not more needs to be done to properly update to click 8.

From the changelog:

> Redesign the shell completion system. [#1484](https://github.com/pallets/click/issues/1484), [#1622](https://github.com/pallets/click/pull/1622)
> 
>     Support Bash >= 4.4, Zsh, and Fish, with the ability for extensions to add support for other shells.
> 
>     Allow commands, groups, parameters, and types to override their completions suggestions.
> 
>     Groups complete the names commands were registered with, which can differ from the name they
>     were created with.
> 
>     The autocompletion parameter for options and arguments is renamed to shell_complete. The
>     function must take ctx, param, incomplete, must do matching rather than return all values, and must
>     return a list of strings or a list of CompletionItem. The old name and behavior is deprecated and will be
>     removed in 8.1.
> 
>     The env var values used to start completion have changed order. The shell now comes first, such
>     as {shell}_source rather than source_{shell}, and is always required.

I've searched for `source_` in Watson's code and found `source_zsh` in `create-completion-script.sh`. I think this is where I bow out and leave things for others with more knowledge and understanding of Watson and Python. Thanks!

MinchinWeb:
Ran into this issue today. Would love to see this merged in.

Xiretza:
It's been almost a year, what's the status on this?

jmaupetit:
Rebased and merged in #471

PR #436, opened 2021-06-05
Add config option for the length of watson log

joelostblom:
I think the default of showing the past 7 days is rather long and usually I only need to see the last few days to know if I need to use `restart` for a previous entry. I modified the code a few months back to show only the last 3 days on my own installation, but figured I might as well make it a config options so others can use it. Personally I don't think there needs to be a new flag added, since there  there are already several flags to control the log command from the CLI, just nothing for the default behavior.


k4nar (commented):
Very good idea :+1: .

Do you think you could add a test for it in `test_cli` ?

joelostblom:
Thanks @k4nar ! I added an attempt to a test. I am not sure whether we need to parameterize the test with multiple dates or if it is enough to just check that the log length is correct since it should be invariant of the current date. When I tried to parameterize with multiple dates I also couldn't figure out how to get `log` to count from the parameterized date since it uses `arrow.now()` internally. Overriding the `now` return value  didn't seem to work and it's late now =) Let me know if you think this is sufficient.

joelostblom:
@k4nar Just a reminder to review this when you get a chance!

joelostblom:
@k4nar Just a ping to review this when you get a chance!

PR #437, opened 2021-06-06
Add gap/no-gap options to restart command

joelostblom:
I often find myself typing in `watson restart -2 -G` or similar and being surprised when it doesn't work. Overall,  I think `restart` should work as similarly as possible to `start`, so I suggest adding the gap options in this PR. 

I also added a generic test for `restart` that can keep track of whether its basic functionality break in future PRs, although it doesn't test the gap options per se (same as for the start test).

I also changed the parameter name from `frame` to `id` to make `restart` consistent with `remove` and `edit`. Since you can pass a frame either by position or id, this param name is not ideal, but neither is `frame` since that is used as a variable name for an actual frame object. This is the smallest  change that brings consistency across commands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- The restart command now accepts the `--gap/--no-gap` options.
+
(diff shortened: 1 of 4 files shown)


jmaupetit (approved):
Looks good to me, thanks! I'll wait for @k4nar validation on this (if he has time to take a quick look).

k4nar (approved):
LGTM :+1:

PR #438, opened 2021-06-06
Add some tips on effective watson usage

joelostblom:
I think it would be helpful to expand on the "getting started" section with some more pointers for how to be effective with watson. The ones suggested in this PR are obviously colored by how I use watson and I am happy for them to be replaced with something more general, or maybe better: for there to be a section where users can share their specific workflows (although that might be more hidden for people going to the website). 

I also fixed a typo I noticed when reading the docs.

### DIFF
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -638,7 +638,7 @@ in the future.
 Example:
 
 
-    $ watson start --at 13:37
+    $ watson start appollo11 --at 13:37
     Starting project apollo11 at 13:37
     
 If the `--no-gap` flag is given, the start time of the new project is set
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -230,7 +230,7 @@ def start(ctx, watson, confirm_new_project, confirm_new_tag, args, at_,
     Example:
 
     \b
-    $ watson start --at 13:37
+    $ watson start appollo11 --at 13:37
     Starting project apollo11 at 13:37
 
     If the `--no-gap` flag is given, the start time of the new project is set
--- a/docs/index.md
+++ b/docs/index.md
@@ -138,5 +138,55 @@ To list all available commands, either [explore the commands documentation](user
 ```bash
 $ watson help
 ```
+`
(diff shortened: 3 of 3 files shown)


docs/user-guide/commands.md
```diff
@@ -638,7 +638,7 @@ in the future.
 Example:
 
 
-    $ watson start --at 13:37
+    $ watson start appollo11 --at 13:37
```
jmaupetit, line 641:
```suggestion
    $ watson start apollo11 --at 13:37
```

watson/cli.py
```diff
@@ -230,7 +230,7 @@ def start(ctx, watson, confirm_new_project, confirm_new_tag, args, at_,
     Example:
 
     \b
-    $ watson start --at 13:37
+    $ watson start appollo11 --at 13:37
```
jmaupetit, line 233:
```suggestion
    $ watson start apollo11 --at 13:37
```

jmaupetit (commented):
Thank you for this @joelostblom! :pray: 

I find it cool but a little bit too opinionated :wink: Maybe we should add a new "real world use cases" section in the documentation to introduce examples of Watson usage?

joelostblom:
Haha yes, it is definitely opinionated =) I like the idea of real world usage scenarios, that's what I was trying to get at with my second suggestion. Where do you want that section? As a third entry in the "User guide" after "Commands" and "Configuration"? 

I can add my  story for now and change it to a more personal narrative and encourage others to share their tips/setup/workflows. I feel like this sharing of knowledge could be good for community building too, not sure about the exact format, but something like this:

> These are examples from the community of how people use watson for their daily time tracking. Please feel free to share your own tips via a GitHub issue and we can add them here!
> 
> User 1/Example 1 (or name?)
>  ...
> User2/Example 2
> .... etc

jmaupetit:
> As a third entry in the "User guide" after "Commands" and "Configuration"?

Yes, good idea!

Few years ago, I wrote a narrative about Watson usage: https://tailordev.fr/blog/2016/02/05/a-day-tracking-my-time-with-watson/

Not sure it's still relevant now...

PR #445, opened 2021-07-17
Feat: add a --compact flag to the status command

EmileRolley:
> Related project scope: `cli`

## Description

In order to use watson output with tools like [starship](https://starship.rs/), this PR add a compact flag (`-c` or `--compact`) to the `status` command.

**Usage:**:
```
> watson status (-c | --compact)
<project> <time_elapsed_in_minutes>m
```

**Screenshot:**

Screenshot from 2021-07-17 18-09-07

### Features

* **cli:** add a --compact flag for the status command ([a16fa48](https://github.com/EmileRolley/Watson/commit/a16fa4800947fb6e72f9235129f5357516b2f5fe))

### DIFF
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -692,6 +692,7 @@ Example:
 
 Flag | Help
 -----|-----
+`-c, --compact` | only output project and time elapsed in a compact format
 `-p, --project` | only output project
 `-t, --tags` | only show tags
 `-e, --elapsed` | only show time elapsed
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -407,6 +407,8 @@ def cancel(watson):
 
 
 @cli.command()
+@click.option('-c', '--compact', is_flag=True,
+              help="only output project and time elapsed in a compact format")
 @click.option('-p', '--project', is_flag=True,
               help="only output project")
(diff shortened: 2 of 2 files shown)


jmaupetit (commented):
What about piping watson's output to `sed` instead?

EmileRolley:
> What about piping watson's output to `sed` instead?

This could be a solution, but it's easier this way.

jmaupetit:
> Do you think that this flag has no relevance @jmaupetit?

I think it's too specific and I try hard to keep Watson small and focused :wink: That being said, if there is a need for this feature, the output format should be configurable IMO. WDYT?

EmileRolley:
> I think it's too specific and I try hard to keep Watson small and focused  That being said, if there is a need for this feature, the output format should be configurable IMO. WDYT?

I understand. By configurable, you mean something more generic such as `watson status --format "%project %elapsed_in_min"`?

jmaupetit:
Yes, kind of. With a jinja-like syntax instead:

```
$ watson status --format "{{ project }} {{ elapsed_time }}"
```

One should define what should be in the context to "compile" the format template.

EmileRolley:
So I close this PR and open an issue to add the `--format` flag?

jmaupetit:
That would be great yes! Thanks!

PR #447, opened 2021-08-01
Fix tag alignment in plain text output of report and aggregate command

tokas:


jmaupetit (commented):
Great, thank you for this! :pray: Can you add a test so that there won't be future regressions?

k4nar (approved):
:+1:

PR #449, opened 2021-08-16
Use the mock standard module

BoboTiG:
It is available as `unittest.mock` since Python 3.3.


k4nar (approved):
Indeed, now that we are Python 3 only I think we can use it :+1: .

BoboTiG:
Gentle ping :)

jmaupetit:
Thank you @BoboTiG!

Issue #444, opened 2021-07-05
Activities extending on several days


ericguirbal:
I've just noticed an inconsistency between the log and aggregate/report  commands for activities extending on several days.

```zsh
> watson add --from "2021-07-04 23:00" --to "2021-07-05 01:00"  textbook
Adding project textbook, started 18 hours ago and stopped 16 hours ago. (id: 507617a)

> watson log --from "2021-07-04"
Sunday 04 July 2021 (2h 00m 00s)
	507617a  23:00 to 01:00   2h 00m 00s  textbook

> watson aggregate --from "2021-07-04"
Sun 04 July 2021 - 59m 59s            
  textbook - 59m 59s

Mon 05 July 2021 - 1h 00m 00s
  textbook - 1h 00m 00s
```

What about an option `day_start` ?

PR #457, opened 2021-12-12
Support for night owls: starting a day at later hour, e.g. 6, to properly report late night work

sfider:
I like to work late at night, even after midnight. Using Watson in such a setting means reports will count part of my work as being done on the next day, while it was done on the same day from my perspective. I've added 'day_start_hour' option to shift time spans used for reports/aggregates to account for me working late.


ericguirbal:
This addresses #444. Thanks.

PR #482, opened 2022-07-15
add interrupt command support

### LINKED ISSUE #470: watson interrupt  (feature request)
Let's say I am working on project A from 09:00, and suddenly at 09:30 I receive an unexpected phone call about project B. The phone call lasts until 09:50, and then I resume my work on project A.

Typically after the call at 09:50, I would correct my time tracking by doing something like:
```bash
watson stop --at 09:30
watson start --at 09:30 projectB
watson stop 
watson start projectA
```

My feature request is to add a new command, **watson interrupt**, that would behave like this: 
```bash
watson interrupt projectB --at 09:30
> Added projectB from 09:30 to now. Restarting projectA.
```

EdgyEdgemond:
closes #470

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -222,6 +222,42 @@ def test_start_existing_frame_stopped(runner, watson, mocker):
     assert watson.current["project"] == "b-project"
 
 
+# watson interrupt existing task with new task, restart existing
+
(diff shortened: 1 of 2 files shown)


jmaupetit:
Thanks! I'll have a look during this week.

PR #481, opened 2022-07-15
support adding frames with offsets

EdgyNortal:
closes #475 

@jmaupetit This ticket looked useful to me, so took a run at this as well.

Issue #475, opened 2022-06-15
Allow relative values for `--to` and `--from` in `watson add`


thomaseizinger:
Usecase: I just did ~30min of work for a project but forgot to track them via watson. Making up the timestamps for `--to` and `--from` is a bit tedious. It would be great if the following command worked:

```
watson add --from "30min ago" --to "now" my-project +my-tag
```

Issue #10, opened 2015-03-30
Improve input date parsing


jmaupetit:
When watson fails at parsing an input date, we must raise a clear message.

k4nar:
Actually I don't think that our current format (`ỲYYY-MM-DD`, which is arrow's default one) is very relevant. Maybe we should think about something more flexible or natural?

willdurand:
> Maybe we should think about something more flexible or natural?

Any examples maybe?

k4nar:
For example git is pretty versatile, it accepts things like "yesterday", "1 week ago" or "15/04/2015" for `git log --since`. I can't find the reference anywhere though.

k4nar:
[parsedatetime](https://github.com/bear/parsedatetime) might be a good option (thx @jmaupetit).

k4nar:
While parsedatetime is great, it's not a perfect fit for Watson. For example it seems to always be expecting MM/DD dates instead of DD/MM. Also it guesses dates in the future (`monday` is equivalent to `next monday`, not `last monday`) but this is not what's expected in the context of Watson.

Right now our system is more restrictive and cumbersome, but at least it's not bug prone.

FTR we must set `YearParseStyle` to 0 in order to parse `Jan 1th` as the current year and not the next.

SpotlightKid:
There's a new (natural language) date parsing library here:

https://github.com/scrapinghub/dateparser

(Haven't tested it yet.)

jmaupetit:
:+1: let's give it a try!

SpotlightKid:
I experimented with `dateparser`a bit. I'm not so sure, I like it. It promises more that it can keep. All below are valid German date specifications. In German, for numbers up to 12, we generally use the word form, not the numeral in a sentence. Also "12 Uhr" normally means mid-day, not midnight, for that we use "Null Uhr" or "0 Uhr" to make it clear.

Not sure if it's worth to add such a heavy dependency when in effect it doesn't add much over `dateutil` .

```
In [1]: import dateparser

In [2]: dateparser.parse('Vor zwei stunden')

In [3]: dateparser.parse('Vor 2 stunden')
Out[3]: datetime.datetime(2015, 11, 14, 0, 4, 24, 713893)

In [4]: dateparser.parse('In 1 Tag')

In [5]: dateparser.parse('Morgen')

In [6]: dateparser.parse('morgen')

In [7]: dateparser.parse('morgen früh')

In [8]: dateparser.parse('morgen mittag')

[...]

PR #484, opened 2022-08-19
Add support for relative dates using dateparser

MatthewWilkes:
Hello! I appreciate this is a long-standing community discussion, and this PR may not be super helpful, but I implemented it to scratch my own itch and thought I'd offer it up.

Following the discussion in #10 there was a proposal to move to using `dateparser` for formatting dates. This isn't a perfect library, it currently makes some silly mistakes like parsing "1.5 hours ago" as "5 hours ago" and can be somewhat slow for confusing formats, but for all other day-to-day requirements I personally have, it handles them well.

This changes the `DateTime` parameter value to fall back on `dateparser` for any dates previously not allowable. This means that all currently acceptable dates are parsed in the same way, but some previously invalid dates are now allowable. 

There's also an issue at #475 for relative dates specifically, which is implemented by PR #481. That implements a subset of the parsing directly using arrow, which I understand may well be preferable. I'm under no illusions that 1 hour of hacking is anything special, but wanted to share rather than hide it away.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,5 @@ arrow>=1.0.0
 click>=8.0
 click-didyoumean
 colorama; sys_platform == "win32"
+dateparser
 requests
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -43,7 +43,8 @@ Else, print the total for each root project.
 
 By default, the time spent the last 7 days is printed. This timespan
 can be controlled with the `--from` and `--to` arguments. The dates
-must have the format `YEAR-MONTH-DAY`, like: `2014-05-19`.
+should have the format `YEAR-MONTH-DAY`, like: `2014-05-19`, but relative
+formats like `two weeks ago` are also supported.
 
 You can limit the report to a project or a tag using the `--project` and
 `--tag` options. They can be specified several times each to add multiple
@@ -234,7 +235,8 @@ Display each recorded session during the given timespan.
 
 By default, the sessions from the last 7 days are printed. This timespan
 can be controlled with the `--from` and `--to` arguments. The dates
-must have the format `YEAR-MONTH-DAY`, like: `2014-05-19`.
+should have the format `YEAR-MONTH-DAY`, like: `2014-05-19`, but relative
+formats like `two weeks ago` are also supported.
 
 You can also use special shortcut options for easier timespan control:
 `--day` sets the log timespan to the current day (beginning at `00:00h`)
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -2,4 +2,5 @@ arrow>=1.0.0
 click>=8.0
 click-didyoumean
 colorama; sys_platform == "win32"
+dateparser
```
EdgyNortal, line 5: given that arrow is already here, wouldn't hurt to rework this change with `arrow.dehumanize(s)`.

Can't talk for the maintainers, but I'd say using existing dependencies rather than adding a new one would increase chance of this being pulled in.

EdgyNortal: This also looks like it could replace my PR more cleanly, so if you were to ingest the tests that my PR supports and take it further to days/weeks etc, that'd be an improvement.

kuon:
Would this support syntax like `-3h` ? This is the syntax I would like the most.

joelostblom:
I implemented a version of this locally where the base (and only) unit was minutes, so you could do do `watson start --at -5` to start 5 minutes ago. I think that worked really well; here is the code in case it is helpful:

```diff
commit ea0c189b1b0a0504638e18efcb1584ef41743af3 (HEAD -> log-length_and_master_merged)
Author: Joel Ostblom <joel.ostblom@gmail.com>
Date:   Mon Feb 28 09:45:43 2022 -0800
 
    Allow negative integers to --at for specifying "X min ago"
 
diff --git a/watson/cli.py b/watson/cli.py
index 580e0af..a21e08c 100644
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -109,20 +109,29 @@ class DateTimeParamType(click.ParamType):
 
     def _parse_multiformat(self, value) -> arrow:
         date = None
-        for fmt in (None, 'HH:mm:ss', 'HH:mm'):
+
+        # Only allow for negative times
+        # to avoid that they are mistaken for the hour of the day
+        if isinstance(value, str) and value[0] == '-':
             try:
-                if fmt is None:
-                    date = arrow.get(value)
-                else:
-                    date = arrow.get(value, fmt)
-                    date = arrow.now().replace(
-                        hour=date.hour,
-                        minute=date.minute,
-                        second=date.second
-                    )
-                break
+                date = arrow.now().shift(minutes=int(value))
             except (ValueError, TypeError):
                 pass
+        else:
+            for fmt in (None, 'HH:mm:ss', 'HH:mm'):
+                try:
+                    if fmt is None:
+                        date = arrow.get(value)
+                    else:
+                        date = arrow.get(value, fmt)
+                        date = arrow.now().replace(
+                            hour=date.hour,
+                            minute=date.minute,
+                            second=date.second
+                        )
+                    break
+                except (ValueError, TypeError):
+                    pass
         return date
```

PR #488, opened 2022-11-30
Adding formats to configuration, closes #94

### LINKED ISSUE #94: [enhancement] color settings
It would be nice to have some control over the color codes being used. I am having trouble reading tags on a windows machine using putty with default color settings (black background).

If I can help out make this happen, in any way, give me hint :smile:

mapio:
This PR adds a `formats` configuration section that can contain configuration for colors (and other) used by click.style.

### DIFF
--- a/watson/watson.py
+++ b/watson/watson.py
@@ -10,7 +10,7 @@ import click
 
 from .config import ConfigParser
 from .frames import Frames
-from .utils import deduplicate, make_json_writer, safe_save, sorted_groupby
+from .utils import deduplicate, make_json_writer, safe_save, sorted_groupby, style
 from .version import version as __version__  # noqa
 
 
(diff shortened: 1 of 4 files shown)


watson/watson.py
```diff
@@ -581,3 +581,9 @@ def rename_tag(self, old_tag, new_tag):
 
         self.frames.changed = True
         self.save()
+    
+    def style(self = None, name = 'error', element = None):
+      if isinstance(self, Watson):
```
mapio, line 586: This is an hack, I should have probably commented it in the code. Here is the rationale.

Having `self` default to `None` is a way to have both (sort of) a class method (used in case the object has not yet been instantiated, for instance by exceptions in `cli`) and an instance method.

If called as `Watson.style` this will not attempt to look at the `config` property of the uninstantiated object; on the other hand, if `watson` is an instance of `Watson`, then `watson.style` will be able to access the configured formats.

PR #495, opened 2023-02-06
Sync via git repository

lkiesow:
This patch allows Watson to sync via a simple git repository, making it unnecessary to run a specific backend. Instead, any git repository with write access will do.

You can configure a git backend by setting something like:

    watson config backend.repo git@github.com:user/repo.git

If y repository is set, the `sync` command will try using git. If not, the old backend server is used if `server` and `token` are set.

### DIFF
--- a/docs/user-guide/commands.md
+++ b/docs/user-guide/commands.md
@@ -739,6 +739,9 @@ Example:
 
     $ watson config backend.url http://localhost:4242
     $ watson config backend.token 7e329263e329
+    or
+    $ watson config backend.repo git@github.com:user/repo.git
+    
     $ watson sync
     Received 42 frames from the server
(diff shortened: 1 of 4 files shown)


pylipp:
@lkiesow I love this! What is your experience using it?

lkiesow:
I've been using this ever since I created this pull request with no major problems.
I've also added this to [watson-next](https://github.com/lkiesow/watson-next) which contains a few additional patches.

pylipp:
> I've been using this ever since I created this pull request with no major problems. I've also added this to [watson-next](https://github.com/lkiesow/watson-next) which contains a few additional patches.

That's great to hear. Have you shared this fork elsewhere? Might be interesting because the project looks for new maintainers and such.

PR #513, opened 2025-11-13
Fixes for Python 3.14

alsmola:
## Summary
- Fixes compatibility issues for Python 3.14

## Changes
- Updated watson/cli.py to address Python 3.14 compatibility

## Test plan
- [ ] Verify Watson works correctly on Python 3.14
- [ ] Ensure backward compatibility with previous Python versions
- [ ] Run existing test suite

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/watson/cli.py
+++ b/watson/cli.py
@@ -666,6 +666,10 @@ def report(watson, current, from_, to, projects, tags, ignore_projects,
                            luna=luna, all=all,
                            include_partial_frames=include_partial_frames)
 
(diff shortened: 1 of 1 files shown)


philpagel:
Thanks for trying to fix this, but I think, the problem is in `class MutuallyExclusiveOption` or the respective `click.option` definitions and we should fix that instead of handling the problem later in the code.

PR #516, opened 2025-12-15
✅(project) fix failing tests 

jmaupetit:
## Purpose

Datetime parsing tests are broken.

## Proposal

- [x] remove frame datetime to local time convertion


jmaupetit:
Depends on #514

